PR #3, opened 2021-07-27
LoadingTextResource: change return type to Nothing

hoc081098:


gabrielittner (approved):
Thanks for the contribution 👍

PR #8, opened 2021-08-23
Add ComposeDialogFragment annotation

kboyarshinov:
This works the same as the already existing `@ComposeFragment` annotation. Similar to `@RendererDialogFragment`  it has new `val dialogFragmentBaseClass: KClass<out DialogFragment>` parameter to support dialogs and bottom sheet with Compose UI in codegen.


whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/WhetstoneCodeGenerator.kt
```diff
@@ -158,6 +164,36 @@ class WhetstoneCodeGenerator : CodeGenerator {
...
+        declaration: KtDeclaration
+    ): GeneratedFile? {
+        val composeFragment = declaration.findAnnotation(composeDialogFragmentFqName, module) ?: return null
+        val data = ComposeFragmentData(
+            baseName = declaration.name!!,
+            packageName = declaration.containingKtFile.packageFqName.pathSegments().joinToString(separator = "."),
+            scope = composeFragment.requireClassArgument("scope", 0, module),
+            parentScope = composeFragment.requireClassArgument("parentScope", 1, module),
+            dependencies = composeFragment.requireClassArgument("dependencies", 2, module),
+            fragmentBaseClass = composeFragment.requireClassArgument("dialogFragmentBaseClass", 3, module),
```
gabrielittner, line 179: I haven't found a way so far to read the default value of an annotation, so for now you need too use `optionalClassArgument` and add `?: dialogFragment`

gabrielittner (approved):

PR #11, opened 2021-09-02
Add apple silicon targets

### LINKED ISSUE #10: Add apple silicon targets
👋  I'm trying [add apple silicon targets for FlowRedux](https://github.com/freeletics/FlowRedux/issues/199) and realized this library is a dependency, so we'd need to update to Kotlin 1.5.30 and Coroutines 1.5.2 and enable apple silicon targets here.

ychescale9:
Fixes #10.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
     }
     dependencies {
         classpath "com.android.tools.build:gradle:7.0.0"
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.21"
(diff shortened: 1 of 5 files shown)


gabrielittner:
I don't want to update the compose beta at this point. I've created a `0.2.1` release that only updates Kotlin for the state machine module. https://github.com/freeletics/mad/commit/a0fb2442f62e636dc6720124fb229f82f190a00a 

This should be enough to unblock FlowRedux and we will update main once a stable compose release supports 1.5.30

PR #19, opened 2022-01-21
add NavEventNavigator implementation

gabrielittner:
This provides an implementation of `Navigator` and `NavEventNavigationHandler` using the concept of `NavEvent`. 

Differences to our internal implementation:
- A `Channel` that is exposed as a `Flow` is used to make the events Observable. Compared to `LiveData` this is not multicasting so there is no need for a "single live event" anymore and the events are simple classes. It also means that tests don't need to use `InstantTaskExecutorRule` anymore.
- No DeepLink support as those are still very tied to our app
- A separate `FragmentNavEventNavigator` that contains the fragment result APIs. This way the general implementation stays "pure" and we don't expose fragments to compose only usages.

A README that I will add separately is still missing but the docs in the code are complete.

### DIFF
--- a/navigator/runtime-compose/build.gradle
+++ b/navigator/runtime-compose/build.gradle
@@ -1,0 +1,58 @@
+apply plugin: 'com.android.library'
+apply plugin: 'org.jetbrains.kotlin.android'
+apply plugin: 'com.vanniktech.maven.publish'
+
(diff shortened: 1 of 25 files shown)


kboyarshinov (approved):
LGTM :+1:

PR #21, opened 2022-01-28
add multi backstack support to nav-events

gabrielittner:
Adds a new `TabRoute` interface to represents routes to a tab. This can then be used with `navigateToTab` which triggers a `NavigateToTab` event. The event will be handled by navigating to the destination using nav options with `popUpTo(..., saveState = true)` and `restoreState(...)`, leveraging the multi backstack support that was added in AndroidX Navigation 2.4

### DIFF
--- a/navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavEventNavigationHandler.kt
+++ b/navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavEventNavigationHandler.kt
@@ -126,12 +126,12 @@ public open class NavEventNavigationHandler : NavigationHandler<NavEventNavigato
                     event.options,
                 )
             }
-            is NavEvent.NavigateToTabEvent -> {
+            is NavEvent.NavigateToRootEvent -> {
                 val options = NavOptions.Builder()
                     // save the state of the current root before leaving it
                     .setPopUpTo(controller.graph.startDestinationId, inclusive = false, saveState = true)
                     // restoring the state of the target root
-                    .setRestoreState(event.restoreTabState)
+                    .setRestoreState(event.restoreRootState)
                     // makes sure that if the destination is already on the backstack, it and
                     // everything above it gets removed
                     .setLaunchSingleTop(true)
--- a/navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavEventNavigationHandler.kt
+++ b/navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavEventNavigationHandler.kt
@@ -137,8 +137,8 @@ public open class NavEventNavigationHandler : NavigationHandler<NavEventNavigato
                     .setLaunchSingleTop(true)
                     .build()
                 controller.navigate(
-                    event.route.destinationId,
-                    event.route.getArguments(),
+                    event.root.destinationId,
+                    event.root.getArguments(),
                     options
                 )
             }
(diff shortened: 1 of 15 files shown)


navigator/runtime/src/main/java/com/freeletics/mad/navigator/TabRoute.kt
```diff
@@ -0,0 +1,22 @@
...
+import android.os.Bundle
+
+/**
+ * This is similar to a [NavRoute] but represents the route to the start destination used in
+ * a backstack. When you navigate to a [TabRoute] the current backstack is saved and removed
+ * so that the [TabRoute] is right on top of the start destination.
+ *
+ * [getArguments] can optionally be overridden to provide extra information to the destination.
+ */
+public interface TabRoute {
```
kboyarshinov, line 12: Thinking whether `TabRoute` is good name for this, and if saving/restoring back stack should be optional per Route. For example in case of side drawer navigation back stack won't reset but the rest would work the same way as tabs do.

gabrielittner: Yes, it doesn't need to be a tab. It's the route to the root of a backstack but `BackstackRootRoute` sounds weird. Do you have any ideas?

kboyarshinov: What about just `NavRoot` and omit `Route` ending?

gabrielittner: I like it 👍

gabrielittner: And updated 2c293e8

navigator/runtime/src/test/java/com/freeletics/mad/navigator/NavEventNavigatorTest.kt
```diff
@@ -45,6 +46,19 @@ public class NavEventNavigatorTest {
         }
     }
 
+    @Test
+    public fun `navigateToTab event is received`(): Unit = runBlocking {
```
kboyarshinov, line 50:
```suggestion
    public fun `navigateToRoot event is received`(): Unit = runBlocking {
```

navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavEventNavigationHandler.kt
```diff
@@ -111,18 +112,34 @@ public open class NavEventNavigationHandler : NavigationHandler<NavEventNavigato
...
-                    navEvent.navRoute.getArguments(),
-                    navEvent.navOptions,
+                    event.route.destinationId,
+                    event.route.getArguments(),
+                    event.options,
+                )
+            }
+            is NavEvent.NavigateToTabEvent -> {
+                val options = NavOptions.Builder()
+                    // save the state of the current tab before leaving it
```
kboyarshinov, line 131:
```suggestion
                    // save the state of the current root before leaving it
```

navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavEventNavigationHandler.kt
```diff
@@ -111,18 +112,34 @@ public open class NavEventNavigationHandler : NavigationHandler<NavEventNavigato
...
+                    .setPopUpTo(controller.graph.startDestinationId, inclusive = false, saveState = true)
+                    // restoring the state of the target tab
```
kboyarshinov, line 133:
```suggestion
                    // restoring the state of the target root
```

kboyarshinov (approved):
LGTM 👍  Couple of renaming suggestions 👇

PR #22, opened 2022-01-30
update NavEntryComponent implementation

### LINKED ISSUE #16: Whetstone: NavEntryComponent should not use string as map key
The class that is passed to `NavEntryId` is not used directly but as a string. This means that when a marker class gets obfuscated the code breaks.

gabrielittner:
- Added `NavEntryComponents` as higher level API to retrieve a generated `NavEntryComponent`. It abstracts away the `Map` of component getters.
- `NavEntryComponentGetter` is now marked as internal API, use `NavEntryComponents` instead
- `NavEntryComponentGetter` will now use the class as map key instead of the fully qualified
name as string. This resolves issues with proguard obfuscation. The only downside is that
`NavEntryComponents` needs to do a one time mapping from `Class` to `String` during construction.

Closes #16


whetstone/runtime/src/main/java/com/freeletics/mad/whetstone/internal/NavEntryComponentGetter.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.freeletics.mad.whetstone.NavEntryComponent
+import com.freeletics.mad.whetstone.NavEntryId
+import dagger.MapKey
+import kotlin.reflect.KClass
+
+/**
+ * A generated implementation of this can be used to retrieve a generated [NavEntryComponent].
+ *
+ * The implementation will be bound into a `Map<String, NavEntryComponentGetter>` were the key
+ * is the same scope that is used in [NavEntryId] and [NavEntryComponent].
```
gabrielittner, line 14:
```suggestion
 * The implementation will be bound into a `Map<Class<*>, NavEntryComponentGetter>` were the key
 * is the same scope that is used in [NavEntryId] and [NavEntryComponent]. It can be used through
 * [com.freeletics.mad.whetstone.NavEntryComponents].
```

kboyarshinov (approved):
👍

PR #23, opened 2022-02-02
split fragment NavEventNavigationHandler to support plain NavEventNavigator

gabrielittner:
- `NavEventNavigationHandler` now only takes a `NavEventNavigator`
- moved `FragmentNavEventNavigator` related parts to `FragmentNavEventNavigationHandler`
- to make this extensibility possible there is now a generic parameter on these classes
- added `Fragment` parameter to `handleEvent` to allow full handling of all events
- applied the last 2 changes also to the compose version of `NavEventNavigationHandler` for parity

This makes migrating to compose easier because you first replace `FragmentNavEventNavigator` with `NavEventNavigator` while still staying in fragment land and then the migration to compose only needs to swap to the compose navigation handler which is an import change.


gabrielittner:
Rethinking this a bit

PR #24, opened 2022-02-03
add NavDestination and graph setup methods

gabrielittner:
This defines nav destinations for both fragments and compose screens that can be used to build a graph without xml. Currently a destination has both a `KClass<NavRoute>` and a `destinationId` as identifier, the latter will probably removed. The methods to build a graph from destinations have a `destinationCreator: (NavDestination) -> AndroidXNavDestination?` that can be used to define custom destinations. We can use that internally to still support our old destinations for a bit.


kboyarshinov (approved):
:+1:

PR #25, opened 2022-02-04
add internal composition local for NavController

gabrielittner:
With this we don't need to pass a `NavController` to the generated composables anymore. Instead our `NavHost` method will provide it through an internal composition local that the navigation handler can then find. One side effect is that it's now required to create the graph for compose navigation through our `NavHost` composables, but since that provides extension functionality I don't see that as a problem.

The screen composables now get a `Bundle` with the arguments for the destination instead. The nice thing here is that we also don't rely on `navController.currentBackStackEntry` to retrieve them anymore. That method generally works fine but there are some edge cases were our screen isn't the current entry. For pure compose we get the arguments from the `NavBackStackEntry` that is passed to us from the navigation library when the destination gets created on the backstack.  Composables inside Fragments directly get the `Fragment` arguments. I think we might be able to go one step further in the future and directly replace the `Bundle` with the `NavDirections` class but for that we need the type info in whetstone. I'll look into that when I also look into generating the provides for `NavDestination`.

### DIFF
--- a/navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavHost.kt
+++ b/navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavHost.kt
@@ -7,6 +7,8 @@ import androidx.navigation.ActivityNavigator
 import androidx.navigation.NavController
 import androidx.navigation.NavDestination as AndroidXNavDestination
 import androidx.navigation.compose.NavHost as AndroidXNavHost
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.staticCompositionLocalOf
 import androidx.navigation.NavGraphBuilder
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.ComposeNavigator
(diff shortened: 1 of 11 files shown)


navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavHost.kt
```diff
@@ -29,40 +31,34 @@ import com.google.accompanist.navigation.material.rememberBottomSheetNavigator
...
 public fun NavHost(
     startRoot: NavRoot,
     destinations: Set<NavDestination>,
-    destinationCreator: (NavDestination) -> AndroidXNavDestination? = { null },
 ) {
     val bottomSheetNavigator = rememberBottomSheetNavigator()
     val navController = rememberNavController(bottomSheetNavigator)
     val startDestinationId = startRoot.destinationId
-    NavHost(navController, startDestinationId, destinations, destinationCreator)
+    NavHost(navController, startDestinationId, destinations) { null }
```
gabrielittner, line 44: The support for custom destinations only makes sense for the overloads below where a `NavHostController` is passed to us. With the one we are creating here it's not possible to create the custom destinations.

kboyarshinov (approved):
:shipit:

PR #26, opened 2022-02-04
simplify Compose/RendererFragment annotations by removing extra dialog versions

gabrielittner:
Replaced the `ComposeDialogFragment` and `RendererDialogFragment` annotations by adding `fragmentBaseClass` to `ComposeFragment` and `RendererFragment` to reduce the API surface and code paths we have. `fragmentBaseClass` defaults to `Fragment` so for all dialogs you need to explicitly set it.


whetstone/runtime-fragment/src/main/java/com/freeletics/mad/whetstone/ComposeFragment.kt
```diff
@@ -1,5 +1,7 @@
 package com.freeletics.mad.whetstone
 
+import androidx.fragment.app.DialogFragment
```
kboyarshinov, line 3: unused import

kboyarshinov (approved):
LGTM :+1:

Would make sense to update `ComposeFragment` and `RendererFragment` documentation with 

```
Generated Fragment will use the given [fragmentBaseClass] as its super class with [androidx.fragment.app.Fragment] class used as default if nothing is specified.
```

gabrielittner:
Added the comment 👍

PR #27, opened 2022-02-04
remove NavOptions from navigateTo

gabrielittner:
Instead of accepting a full `NavOptions` object we only take the `popUpTo` id and whether it's inclusive. The motivation here are 2 things:
- together with #25 we limit exposing `androidx.navigation` APIs to creating the graphs (`NavHost` composable and `NavHostFragment.setGraph(...)`)
- when we remove `destinationId` we need to replace the `popUpTo` arg and if our API takes an already built `NavOptions` object that is not possible

Currently we don't use any other available options so I limited to these but we could add more if needed.


kboyarshinov (approved):
Limit exposing `androidx.navigation` makes sense to me. :+1:

PR #28, opened 2022-02-04
remove ResultLauncher interface

gabrielittner:
Makes the activity result and permission result launchers independent. It was a bit confusing that they shared the same base interface. While it allowed us to share some code on the event handling side, it resulted in a confusing public API. Our navigator has `navigateForResult` which took the now removed interface and it has `requestPermissions`. For permission results you want to call the latter but the API allowed you to call both. With this it's split you can only call `requestPermissions`.

Note: I will deduplicate some of the `NavEventNavigationHandler` code in a follow up pr.


kboyarshinov (approved):
LGTM :+1:

PR #29, opened 2022-02-04
add missing ModalBottomSheetLayout to NavHost composables

gabrielittner:
The simple NavHost composables needs to do add `ModalBottomSheetLayout` themselves since they are creating the  `bottomSheetNavigator` internally.


kboyarshinov (approved):
:shipit:

PR #30, opened 2022-02-04
share NavEvent handling and permission result logic

gabrielittner:
Removes the duplicate logic that we had in the 2 navigation handlers. With this we can also mark all the fields inside the nav event classes as `internal`. The classes themselves and their constructors stay public so that it is possible to instantiate them to test emitted events by comparison. 

Until https://youtrack.jetbrains.com/issue/KT-22284 is done the shared methods are only marked with `InternalNavigatorApi` and are technically public.

### DIFF
--- a/navigator/runtime-fragment/src/main/java/com/freeletics/mad/navigator/fragment/NavEventNavigationHandler.kt
+++ b/navigator/runtime-fragment/src/main/java/com/freeletics/mad/navigator/fragment/NavEventNavigationHandler.kt
@@ -5,38 +5,29 @@ import android.os.Bundle
 import androidx.activity.result.ActivityResultCaller
 import androidx.activity.result.ActivityResultLauncher
 import androidx.annotation.CallSuper
-import androidx.core.app.ActivityCompat.shouldShowRequestPermissionRationale
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentResultOwner
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.coroutineScope
 import androidx.lifecycle.repeatOnLifecycle
-import androidx.navigation.NavOptions
 import androidx.navigation.fragment.findNavController
 import com.freeletics.mad.navigator.NavEvent
 import com.freeletics.mad.navigator.NavEventNavigator
-import com.freeletics.mad.navigator.NavEvent.BackEvent
-import com.freeletics.mad.navigator.NavEvent.BackToEvent
-import com.freeletics.mad.navigator.NavEvent.NavigateToEvent
-import com.freeletics.mad.navigator.NavEvent.ActivityResultEvent
-import com.freeletics.mad.navigator.NavEvent.UpEvent
-import com.freeletics.mad.navigator.internal.InternalNavigatorApi
 import com.freeletics.mad.navigator.ActivityResultRequest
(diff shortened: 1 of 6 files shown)


navigator/runtime/src/main/java/com/freeletics/mad/navigator/NavEvent.kt
```diff
@@ -16,26 +16,26 @@ public interface NavEvent {
      * Navigates to the given [route].
      */
     public data class NavigateToEvent(
```
kboyarshinov, line 18: Could only comment here. 

`import androidx.annotation.IdRes` is unused.

navigator/runtime-fragment/src/main/java/com/freeletics/mad/navigator/fragment/NavEventNavigationHandler.kt
```diff
@@ -105,68 +88,13 @@ public open class NavEventNavigationHandler : NavigationHandler<FragmentNavEvent
...
-                )
-                @Suppress("UNCHECKED_CAST")
-                (launcher as ActivityResultLauncher<Any?>).launch(event.permissions)
-            }
-            is FragmentResultEvent -> {
-                val result = Bundle(1).apply {
-                    putParcelable(KEY_FRAGMENT_RESULT, event.result)
-                }
-                fragment.parentFragmentManager.setFragmentResult(event.requestKey, result)
+        if (event is FragmentResultEvent) {
```
kboyarshinov, line 91: Should fields of `FragmentResultEvent` be marked as internal too?

kboyarshinov (approved):
LGTM 👍 

With the change to `NavEvent` field being now internal, it would make overridable `handleNavEvent` incapable to handle existing events differently, making it solely for custom events.  Wondering if then it would be better to have different interface for custom navEvents e.g.

```kotlin
interface CustomNavEvent: NavEvent
```

Since otherwise it would be inconvenient to have exhaustive `when(navEvent) { .. } ` that lists all internal nav events that can't be interacted. I guess the only overridable option would be to ignore internal nav events entirely. 

Then we can make library navevents `internal` too and provide testing substitute in another artifact for verifications only. 

Or we can disallow handing of custom nav events entirely since to actually handle any custom event you need to access either `fragment` or `navController`.

What do you think?

navigator/runtime-fragment/src/main/java/com/freeletics/mad/navigator/fragment/NavEventNavigationHandler.kt
gabrielittner: I have a follow up that splits out `FragmentResultEvent` into it's own thing that doesn't extend `NavEvent` anymore. That change also makes the fields internal.

gabrielittner:
> Or we can disallow handing of custom nav events entirely since to actually handle any custom event you need to access either fragment or navController.

That's what i wanted to do after the `FragmentResultEvent` change mentioned above. We don't need custom events in the library anymore then and internally we also have no more use case for it.

navigator/runtime/src/main/java/com/freeletics/mad/navigator/NavEvent.kt
gabrielittner: [acace8d](https://github.com/freeletics/mad/pull/30/commits/acace8ddd54e09c8690e3b4a1e05246cb0336a0c)

PR #31, opened 2022-02-04
move FragmentResultEvent into it's own stream

gabrielittner:
With this `FragmentResultEvent` does not extend `NavEvent` anymore and has it's own `Flow`. Instead of having an extra back event that's being sent the handler for this event will just do both. Since this is a very specific use case there shouldn't be any issues with ordering, delivering a fragment result and going back doesn't make sense to be mixed with other events.

In the next PR I will lock down the APIs that now don't need to be exposed anymore (NavEvent as sealed class, some extension points etc).


kboyarshinov (approved):
:ship:

PR #32, opened 2022-02-07
lock down NavEvent and Navigator APIs

gabrielittner:
Following the previous changes `NavEvent` is now a sealed class. With this the handlers don't need to be `open` anymore and don't need `handleEvent` methods and `sendNavEvent` in the navigator is now private. While going through the public API I've also added `VisibleForTesting` to `NavEvent`, the Flows and other APIs that are only public for test purposes. If we decide on doing an extra testing artifact it's easier to find these.


kboyarshinov (approved):
:+1:

PR #33, opened 2022-02-07
add binary compatibility validator to make reviewing the public api easier

gabrielittner:
For example it's easy to search for `androidx/navigation` to find where we still expose it in the public API. 

I've set it to consider our internal annotations and VisibleForTesting as non public. There is a small bug though that the getters from properties marked as internal are still in the dump https://github.com/Kotlin/binary-compatibility-validator/issues/36


kboyarshinov (approved):
LGTM :+1:

PR #34, opened 2022-02-07
replace custom destination support with default args API that is marked as obsolete

gabrielittner:
We had the `destinationCreator` lambda that allowed turning a custom `NavDestination` into an AndroidX `NavDestination`. The API existed so that we can support our internal custom destination which receives some extra options. Instead of that the destinations now have a `Bundle` parameter to directly pass in these options (we can still have a higher level API that internally builds the bundle in our codebase). That allows us to lock down `NavDestination` by making it a sealed interface and changing the parameters to internal. Together the following pull request it also completely removes AndroidX navigation from the public API for compose. For Fragments the only remaining part is `NavHostFragment` which we can't replace, but it's also less important to hide AndroidX navigation completely in there (I don't think we would even consider swapping the implementation for fragments). 

The constructors of `NavDestination` that receive a `Bundle` are marked with a new `ObsoleteNavigatorApi` annotation to signify that these only exist for now to enable the Freeletics codebase to work, but will be removed at some point.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,6 +51,7 @@ apiValidation {
     nonPublicMarkers += [
         "androidx.annotation.VisibleForTesting",
         "com.freeletics.mad.navigator.internal.InternalNavigatorApi",
+        "com.freeletics.mad.navigator.internal.ObsoleteNavigatorApi",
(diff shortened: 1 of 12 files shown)


kboyarshinov (approved):
:+1:

PR #35, opened 2022-02-07
add an API to obtain the NavController outside of compose (marked as obsolete)

gabrielittner:
With how the nav entry lookup in whetstone currently works we need access to the `NavController` outside of compose. This adds an API that lets you do that for now until we change how whetstone works. It's still a bit hacky but generally this is the same thing that AndroidX navigation does for fragment navigation to enable looking up nav controllers from views and activities. 

AndroidX Navigation is now completely gone from the public api for compose users.


kboyarshinov (approved):
LGTM :+1:

PR #36, opened 2022-02-07
extract methods out of NavEventNavigationHandler classes

gabrielittner:
The logic of the handlers is now available as standalone methods. This allows to integrate the navigation library more nicely without using whetstone (for compose you just throw a `NavigationSetup(navigator)` into your composable instead of instantiating the class and then calling the method on it. At the same time I'm also planning to get rid of the handler classes themselves. They exists so that you can hook other navigation systems into it, but I think we can be more opinionated. The integration stays optional (whetstone and navigator can both be uses without each other) but if you want integrated navigation you'll have to use them together. It's still possible to build custom navigation solutions without having them in the generated code.


kboyarshinov (approved):
LGTM :+1:

PR #38, opened 2022-02-08
remove Parcelable from NavRoute/NavRoot interfaces

gabrielittner:
When navigating to a screen in another app the other app will crash if we put our parcelable into the arguments. Previously we were explicitly overriding `getArguments` but that isn't exposed anymore. So I made the interfaces not parcelable and if a route isn't parcelable we will have an empty `Bundle`.  `NavRoot` doesn't really need this change because it should be something internal, but I wanted to keep it in sync.

I was thinking whether there should be a `ParcelableNavRoute`  but that isn't really shorter than writing `NavRoute, Parcelable`.


kboyarshinov (approved):
:+1:

PR #39, opened 2022-02-08
add @NavDestination for integration with destinations

gabrielittner:
This is the start of a more opinionated navigation integration for Whetstone. Instead of having `navigator` and navigationHandler` parameters on the regular annotation, there is a new `@NavDestination` annotation. The presence of this is used to include the navigation setup in the generated code. Since we don't know the concrete navigator class anymore it's expected that it's bound to `NavEventNavigator` or `FragmentNavEventNavigator` (depending on compose whether we are generating a fragment). 

Next steps:
- remove the 3 `navigator-common` modules they are obsolete now
- add parameters to `@NavDestination` that allow us to use `NavRoute` in generated code and actually generate the `NavDestination`


kboyarshinov (approved):
:shipit:

PR #40, opened 2022-02-09
remove navigation-common

gabrielittner:
Follow up to #39. This removes the navigation-common artifacts and base interfaces because they are not used anymore.


kboyarshinov (approved):
:shipit:

PR #41, opened 2022-02-09
add NavRoute/Root parameters to @NavDestination, eliminate Bundle

gabrielittner:
The `@NavDestination` annotation now gets a `NavRoute` as mandatory parameter. There is also a `RootNavDestination` annotation for `NavRoot`. Whetstone will use this to make that class instead of a `Bundle` available in the generated component. On the compose side this is also what we pass to the generated composable instead of a `Bundle`. When there is no nav annotation the current `Bundle` behavior is kept.

On the navigation side this also allows some changes. The destinations for compose will now directly get an instance of the route passed to them instead of the `Bundle`, so a pure compose user doesn't see at all anymore how we deal with arguments internally. The `Bundle.toNavRoute/Root` methods are now marked as internal instead of obsolete. On the fragment side we have `requireNavRoute/Root` extensions on `Fragment` to obtain them. These are used in the code generated by whetstone, but also public for using the navigation lib without whetstone.


kboyarshinov (approved):
LGTM :+1: I guess `NavDestination` in `:whetstone:runtime` is no longer needed, right?

gabrielittner:
Yes I forgot to delete that 30a58d660a819617be6fc38aa0a4280c9973dfcc

PR #42, opened 2022-02-09
add DestinationType to @NavDestination

gabrielittner:
This will be used in the codegen to provide a `NavDestination` into the given `destinationScope`. The type determines what we generate. `NONE` is a temporary opt out mechanism for cases where a destination with custom default args is needed. Like the default args API it's marked as obsolete.

The codegen itself needs to wait for the destination id removal. I'm opening this now already so that I can already add this info when migrating our codebase to the annotation.


kboyarshinov (approved):
:shipit:

PR #44, opened 2022-02-10
set arguments bundle for the start destination

gabrielittner:
Since our generated code expects the route in the arguments for each screen (for compose it's even the navigator library itself) we also need to set the arguments for the start destination. For Fragments this is easy, we can just pass them when we call the AndroidX `setGraph` method. For compose it's more complicated. Setting the graph on the controller happens inside their `NavHost` composable and that only accepts args in you are using String routes instead of id + Bundle (the args are then already part of the route that you pass). To work around that we add the start destination args as default args of the destination when building the graph, we can match those base on the route class.


kboyarshinov (approved):
:shipit:

PR #47, opened 2022-02-14
add BaseRoute interface to reduce code duplication for NavRoute/Root

gabrielittner:
When I built #46 last week and also another new API I didn't like that there was so much duplication for `NavRoute` and `NavRoot`. This introduces a sealed base interface for them so that we can have methods accepting both.

There are still 2 scenarios were we separate them:
- `navigateTo` only takes `NavRoute` and there is a `navigateToRoot` for `NavRoot`. The former is simply adding a destination to the backstack while the latter switches out the whole backstack, so I think it makes sense to have that explicit. (this is also why `navigateToRoot` isn't just an overload of `navigateTo`)
- the `@NavDestination` annotations are separate to ensure that a `NavRoot` won't have the wrong type assigned to it because it can never be a dialog or bottom sheet.


kboyarshinov (approved):
LGTM :+1:

PR #49, opened 2022-02-14
fix legacy NavController API

gabrielittner:
Turns out that `setTag` only wants ids generated by aapt and not the ones generated at runtime.

### DIFF
--- a/navigator/runtime-compose/src/main/res/values/public.xml
+++ b/navigator/runtime-compose/src/main/res/values/public.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <public type="id" name="id_that_does_not_exist" />
+</resources>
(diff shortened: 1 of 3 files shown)


navigator/runtime-compose/src/main/res/values/public.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <public type="id" name="id_that_does_not_exist" />
```
kboyarshinov, line 3: is it used?

kboyarshinov (approved):

navigator/runtime-compose/src/main/res/values/public.xml
gabrielittner: The last time I tried public.xml it didn't work if the file was just empty. So I'm declaring something as public that doesn't exist so that all real resources are considered private.

PR #50, opened 2022-02-14
make NavEventNavigator open instead of abstract

gabrielittner:
For simple projects it's a lot of boilerplate to always extend the class and it's perfectly fine to just use it as is.


kboyarshinov (approved):
:+1:

PR #52, opened 2022-03-07
update dependencies and snapshot publishing

gabrielittner:
For easier testing I've changed the publish snapshot action to publish from all branches and from prs. It will use either the branch name or the pr number as version.


kboyarshinov (approved):
:+1:

PR #54, opened 2022-03-10
Use @ContributesMultibinding for generated NavEntryComponentGetter 

gabrielittner:
Before we were generating an extra modules that bound `NavEntryComponentGetter` into the map. The only reason for that was a bug in Anvil that caused it to fail when a plugin would generate something with `@ContributesMultibinding`. This bug was fixed in the latest release so we cans simplify our code.


kboyarshinov (approved):
:+1:

PR #55, opened 2022-03-10
Use @ContributesSubcomponent for generated subcomponent instead of @MergeSubcomponent

gabrielittner:
MergeSubcomponent does the merging with contributed modules/interfaces when the subcomponent gets compiled. This means it needs to be defined somewhere where everything is available, which is in many cases the app module. `@ContributesSubcomponent` delays the merging to when the regular component that the subcomponent is added to gets compiled, so we can also define nav entry components in other modules now.


kboyarshinov (approved):
:+1:

PR #58, opened 2022-03-14
make register methods protected

gabrielittner:
Makes it a bit harder to wrongly use the methods by calling them to late. Now you need to call it from a subclass which forces you to put it into the initialization code of it.


kboyarshinov (approved):
:+1:

PR #62, opened 2022-03-16
remove obsolete destination APIs

gabrielittner:
Removes the APIs to pass default arguments to destinations now that we don't use them anymore internally. Also removes the option to not generate a destination because there are no custom destinations anymore.


kboyarshinov (approved):
LGTM :+1:

PR #66, opened 2022-03-17
unify Fragment codegen, deduplicate ComposeFragment code

gabrielittner:
This refactoring has 2 goals without changing any functionality
- For compose fragments we currently obtain the component twice, once in the fragment to setup navigation and then again in the generated composable. It's not really an issue but it's still unnecessary. This splits the generated composable into 2. The second composables gets the component as it's only parameter. For pure compose we will generate both composable for fragments we only generate the second one.
- Unify the generated fragment code and share the common parts. With this fragments for compose and renderer share most of their code and only the last part of `onCreateView` differs


kboyarshinov (approved):
LGTM :+1:

PR #67, opened 2022-03-17
improve naming for NavEntry variables in generated code

gabrielittner:
- changes the name of the method inside `ParentComponent` to `factory` because that is what it returns
- the view model holds the factory instead of the parent component
- improved naming in the code that obtains the view model


kboyarshinov (approved):
:+1:

PR #69, opened 2022-03-17
move rememberViewModelProvider to runtime-compose

gabrielittner:
With the changes from #66 the method is only used by pure compose code so we can move it to the compose module.


kboyarshinov (approved):
:+1:

PR #71, opened 2022-03-17
move @NavEntryComponent to other annotations, use route to get id

gabrielittner:
`NavEntryComponent` didn't get any updates with all the navigation route changes and it still used destination ids. Instead of putting it on an extra function the annoation now lives next to the other on the screen to which the component should be tied to.

```
@ComposeScreen(...)
@NavDestination(...)
@NavEntryComponent(...)
```

It will take the `route` and `destinationScope` from `@NavDestination` to for the code generation. The route is used automatically during the lookup of the component in the generated `NavEntryComponentGetter` and the getters are now all contributed to the destination scope which puts the map into the same component as the set of destinations. This also makes `@NavEntryId` obsolete and it was removed.

Theoretically we could use the parent scope of `@ComposeScreen` automatically as scope for the nav entry component but I like the consistency of the parameters. `destinationScope` could also serve as parent scope here (it would disallow nesting NavEntryComponents but I'm not sure if that even works right now).

Next steps:
- put an instance of the route into the generated component instead of the arguments bundle (like we already do for regular components)
- refactor the lookup so that retrieving a `NavBackStackEntry` happens internally, this would then allow us to remove our last obsolete navigation API (the one that allows to obtain the `NavController` for `getSystemService`)

### DIFF
--- a/whetstone/runtime/src/main/java/com/freeletics/mad/whetstone/NavEntryComponent.kt
+++ b/whetstone/runtime/src/main/java/com/freeletics/mad/whetstone/NavEntryComponent.kt
@@ -4,8 +4,8 @@ import android.content.Context
 import androidx.navigation.NavBackStackEntry
 import com.freeletics.mad.whetstone.internal.NavEntryComponentGetter
 import javax.inject.Inject
-import javax.inject.Qualifier
 import kotlin.annotation.AnnotationRetention.RUNTIME
+import kotlin.annotation.AnnotationTarget.CLASS
 import kotlin.annotation.AnnotationTarget.FUNCTION
 import kotlin.reflect.KClass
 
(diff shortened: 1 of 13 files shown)


whetstone/runtime/src/main/java/com/freeletics/mad/whetstone/NavEntryComponent.kt
```diff
@@ -43,7 +43,7 @@ import kotlin.reflect.KClass
  * It's recommended to put this annotation on the method that provides the matching [NavEntryId]
```
kboyarshinov, line 43: This comment isn't relevant anymore

kboyarshinov (approved):
:+1:

whetstone/runtime/src/main/java/com/freeletics/mad/whetstone/NavEntryComponent.kt
gabrielittner: [4d5631f](https://github.com/freeletics/mad/pull/71/commits/4d5631fae0f23bb55ebe10e6f01674084e88f84e)

PR #73, opened 2022-03-21
refactor dependency lookup

gabrielittner:
- change `ViewModelProvider` methods to methods that return a `ViewModel`
- don't pass parameters to `WhetstoneViewModelFactory` that are then just passed back to us
- dependency lookup now happens in the `viewModel` methods instead of the in the factory

These changes are in preparation of making the nav entry lookup internal without requiring a `NavController` on the outside (and therefor removing the last obsolete navigation api). We will have overloads of the viewModel methods for that.


kboyarshinov (approved):
:+1:

PR #74, opened 2022-03-21
move whetstone nav code to extra modules

gabrielittner:
With this everything that is for integration with the navigator library goes into extra artifacts. It already eliminates the `androidx.navigation:navigation-common` dependecy and also the compileOnly dependencies we had before. With the upcoming refacotring of how we look up nav entry components we would need even more navigation dependencies so I'm splitting them beforehand.

Dependency wise the new navigation artifacts expose the runtime artifacts so there is no need to depend on both you just choose between runtime and navigation based on whether you use navigator or not. (this is what the dependency analysis excludes are for)


kboyarshinov (approved):
👍

PR #75, opened 2022-03-23
nav entry component look up part 1, internal find entry

gabrielittner:
- we are now looking up `NavEntryComponents` itself through `getSystemService` not the components it provides access to, this removes the need to have a `NavController` available in `getSystemService`
- there is a new `findDependencies` that will search for `NavEntryComponents` first and will try to get the parent from there, this method requires passing in something to look up a `NavBackStackEntry`
- there are also overloads for the `viewModel`/`rememberViewModel` methods that use the new `findDependencies` and will obtain the `NavController`
- the existing `viewModel`/`rememberViewModel` will now only take a `Bundle` as extra which is how the overloads are differentiated, screens that don't use whetstones integration with navigator don't need the new lookup
- the nav entry `viewModel` method always uses the new `findDependencies`

Next steps:
- remove obsolete navigator api to look up `NavController`
- Remove `NavEntryComponents` we can use Anvil and our codegen to contribute an equivalent interface to the `destinationScope` component. This component should already be obtainable through `getSystemService` as it is on top of the hierarchy and regular screens would look it up. After this there is 0 extra configuration needed for nav entry component support, just adding the annotation is enough


kboyarshinov (approved):
:+1:

PR #76, opened 2022-03-24
remove old ObsoleteNavigatorApi

gabrielittner:
With #75 we don't need `findComposeNavController` anymore which removes our last API that is marked as obsolete 🎉


kboyarshinov (approved):
:shipit:

PR #77, opened 2022-03-30
replace NavEntryComponents with automatically contributed interface

gabrielittner:
The last part of the nav entry rewrite. The lookup methods now get the destinationScope passed to them and try to lookup `DestinationComponent` using it. The latter is the replacement for `NavEntryComponents` and allows access to the generated component getters. To make the integration seamless each screen where the `destinationScope` does not equal the `parentScope` will generate an interface that extends `DestinationComponent` which will be contributed to destinationScope. This way the component will automatically extend `DestinationComponent` without any manual integration.


kboyarshinov (approved):
LGTM :+1:

PR #78, opened 2022-04-05
move DestinationComponent to the top level

gabrielittner:
The generated `DestinationComponent` interface doesn't really have a relation with the generated component so it doesn't make sense to have it as a nested class.


kboyarshinov (approved):
:+1:

PR #83, opened 2022-04-19
fix DestinationComponent lookup crashing if there is none

gabrielittner:
This crash happens if no screen in an app uses `@NavEntryComponent`. The general dependency lookup will try to lookup `DestinationComponent` in case it holds the dependencies for it and that will crash because the lookup based on `destinationScope` succeeds but the returned component does not extend `DestinationComponent`. The workaround is to manually extend `DestinationComponent` until this fix is out.

Changes
- Our internal `find` that does the `getSystemService` call will not cast anymore and just returns `Any`
- Both the regular `findDependencies` and the `findDependencies` for navigation will first do a `checkNotNull` with a proper error message on what `find` returned, this will generally make debugging issues easier because you're not just getting a random NPE
- Afterwards `findDependencies` will do the cast
- For the navigation `findDependencies` specifically the `find` call for `DestinationComponent` will now not crash anymore and we do a conditional cast with `as?`, if it's not found we will just do a regular dependency lookup. This fixes the core issue.
- As a small optimization if `scope` and `destinationScope` are equal we skip the whole destination component lookup because we know that the dependencies can't come from a custom scope with a `NavEntryComponentGetter`

### DIFF
--- a/whetstone/navigation-compose/src/main/java/com/freeletics/mad/whetstone/compose/internal/NavComposeViewModelProvider.kt
+++ b/whetstone/navigation-compose/src/main/java/com/freeletics/mad/whetstone/compose/internal/NavComposeViewModelProvider.kt
@@ -27,7 +27,7 @@ import kotlin.reflect.KClass
 @InternalWhetstoneApi
 @OptIn(InternalNavigatorApi::class)
 @Composable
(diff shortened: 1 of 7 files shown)


kboyarshinov (approved):
LGTM :+1:

PR #85, opened 2022-04-21
introduce ActivityRoute to enable dynamically building intents

gabrielittner:
Navigating to an internal `Activity` or an external one where you can statically define the full `Intent` in the destination already worked well. However if you need to dynamically add something to the `Intent` like a `data` `Uri` or some extra there was no direct way to do it. The workaround was to use a destination that builds the intent and then removes itself from the backstack after launching it. This worked pretty well with DialogFragments but with compose navigation you see blinking.

To solve it there is now `ActivityRoute` which is a route to activities. It has a `fillInIntent` method that can optionally be overridden to build an `Intent` based on parameters passed to the route. Internally we have a customised and stripped down copy of AndroidX's `ActivityNavigator` that will merge the destination intent and the fillInIntent using `Intnet.fillIn`. I initially didn't use a custom navigator and just allowed building a `Bundle` inside `ActivityRoute`. AndroidX would automatically add the `Bundle` to the launched `Intent` but that is limited to extras and we have mulitple use cases were we set `data` dynamicallly.

`ActivityRoute` is not a `BaseRoute` which means it's now also not possible anymore to accidentally use routes to an activity in places where it doesn't make sense like `navigateBackTo` or in whetstone annotations. The `ActivityNavDestination` enforces that `ActivityRoute` is used for activities. A nice side effect is that we can now make `BaseRoute` extend `Parcelable` by default. Before it would have caused crashes when navigating to external dependencies, not it's not an issue anymore since the hierarchies are separate. It's still possible to manually implement `Parcelable` for `ActivityRoute`s that go internal activities.

Also added a section to the README, I completely forgot to mention activities there.

### DIFF
--- a/navigator/README.md
+++ b/navigator/README.md
@@ -245,10 +245,11 @@ class ShareRoute(
     private val title: String,
     private val message: String
 ) : ActivityRoute {
-    // the returned Bundle is added to the launched `Intent`
-    override fun intentExtras() = Bundle().apply {
-        putString(Intent.EXTRA_TITLE, title)
-        putParcelable(Intent.EXTRA_INTENT, Intent().apply {
+    // the returned Intent is filled into the Intent of the destination by calling
+    // destinationIntent.fillIn(fillInIntent())
+    override fun fillInIntent() = Intent()
+        .putExtra(Intent.EXTRA_TITLE, title)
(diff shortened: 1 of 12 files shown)


kboyarshinov (approved):
LGTM 👍

PR #104, opened 2022-06-09
add Activity.requireRoute API

gabrielittner:
We didn't have any public API that isn't marked as internal to do this until now. Fragments already had one and for compose it's passed directly to your destination.


snappdevelopment (approved):
The method feels a little bit misplaced, but I also can't find a more fitting place in the repo 🤔

gabrielittner:
I'm also not 100% happy with the location. In the navigator-fragment module it's in it's own file called `FragmentRoute`, so I wanted to create `ActivityRoute` but then thought it doesn't make sense if we already have the route file.

PR #109, opened 2022-06-13
add destination change callback to NavHost

gabrielittner:
Allows listening to nav destination changes for debugging purposes or something like updating a bottom navigation view.


kboyarshinov (approved):
:+1:

PR #118, opened 2022-06-20
support injecting a Set<Closeable> into generated ViewModel

gabrielittner:
Changes:
- we generate a module that has a @Multibinds for `Set<Closeable>`
- the component has a `Set<Closeable>` property
- the generated `ViewModel` will obtain the closeables in `onCleared` and close them
- the generated code for `rxJavaEnabled` and `coroutinesEnabled` has been refactored to use this mechanism, the new module will just contain provides for these

In the release after the next one I then want to completely remove `rxJavaEnabled` and `coroutinesEnabled` completely as this mechanism is a more generic replacement. Doing this in 2 releases makes the migration easier for us though. The next release of AndroidX lifecycle also adds a `addCloseable` method to the `ViewModel` so we could use that instead of manually closing it https://developer.android.com/reference/androidx/lifecycle/ViewModel?hl=en#addCloseable(java.io.Closeable) (not sure if it's worth it though since it's the same amount of lines)


kboyarshinov (approved):
LGTM :+1: 

Perhaps few more tests can be added for the case with both `rxJava` and `coroutineScope` generation disabled, since those will be eventually removed anyway.

gabrielittner:
Added a test for that 👍

PR #123, opened 2022-06-27
use qualifiers for everything automatically provided in nav entry component

gabrielittner:
The automatically provided `SavedStateHandle`, `NavRoute`, `CoroutineScope` and `CompositeDisposable` objects now use an `@NavEntry(Scope::class)` qualifier.

Reasoning
- when we generate subcomponents these would clash with the provided classes from the screen level component
- would already happen for nested NavEntryComponents
- generally more clarity on what you're using and where it comes from


kboyarshinov (approved):
:+1:

PR #125, opened 2022-06-29
make the retained component a contributed subcomponent

gabrielittner:
With this Whetstone will only generate subcomponents instead of full components which allows not using kapt on Gradle modules using Whetstone and also removes the need for the dependencies interface. This PR makes the changes to the code generation I will have a follow up to cleanup the annotations and the docs, as well as one follow up to unify the codegen for nav entry components with the general codegen.


kboyarshinov (approved):
👍

PR #139, opened 2022-06-30
updates for AndroidX lifecycle 2.5.0

gabrielittner:
- use the StateFlow based API to collect results from `SavedStateHandle`, this removes the need to depend on live data
- use the new `ViewModel` `CreationExtras` API and `viewModelFactory` DSL to create ViewModels without needing an extra class


kboyarshinov (approved):
👍

PR #141, opened 2022-06-30
make naming of generated code more consistent, merge nav entry code

gabrielittner:
All generated classes are now prefixed with `Whetstone` and are generally more consistent. I also removed some leftovers like calling it `RetainedComponent`. This also enabled me to completely delete the nav entry specific component, view model and module generators while only making very small changes to the regular variants.

Functionality wise the code is identical to before which you can see in the tests, it's just the names that changed.


kboyarshinov (approved):
LGTM 👍

PR #148, opened 2022-07-11
use different placeholder for default result StateFlow

gabrielittner:
When we switched from `LiveData` to `StateFlow` for observing navigation results through a `SavedStateHandle` (#139) we had to introduce a default placeholder because `StateFlow` requires an initial value. This placeholder would then just get filtered out. The issue is that `getStateFlow` internally will set that initial value on the `SavedStateHandle` which means it will get restored from there as well and our reference based equality doesn't work, resulting in a `ClassCastException`. We stay with the same approach but use a custom parcelable as default value so that we can just do a regular equals check that always works.


snappdevelopment (approved):
👍

PR #212, opened 2022-11-07
Use requireNotNull contracts for requireRoute functions

kboyarshinov:


navigator/runtime/src/main/java/com/freeletics/mad/navigator/internal/Navigate.kt
```diff
@@ -87,7 +87,9 @@ public fun KClass<out ActivityRoute>.activityDestinationId(): Int = internalDest
 private fun KClass<*>.internalDestinationId() = qualifiedName!!.hashCode()
 
 @InternalNavigatorApi
-public fun <T : BaseRoute> Bundle.toRoute(): T = getParcelable(EXTRA_ROUTE)!!
+public fun <T : BaseRoute> Bundle?.requireRoute(): T = requireNotNull(this?.getParcelable(EXTRA_ROUTE)) {
+    "Bundle doesn't contain Route data in \"$EXTRA_ROUTE\" or it is null"
+}
```
gabrielittner, line 92: I think it would make sense to split this into to either 2 separate `requireNotNull` here or to make `Bundle` not nullable again and instead already do a requireNotNull on it on the call site. This is already happening kind of in `FragmentRoute.kt` since were are using `requireArguments()` there.

navigator/runtime/src/main/java/com/freeletics/mad/navigator/internal/Navigate.kt
```diff
@@ -87,7 +87,9 @@ public fun KClass<out ActivityRoute>.activityDestinationId(): Int = internalDest
...
 
 @InternalNavigatorApi
 public fun <T : ActivityRoute> Bundle.toActivityRoute(): T = getParcelable(EXTRA_ROUTE)!!
```
gabrielittner, line 95: Should this also use requireNotNull?

kboyarshinov: No, in next PR I'll change this to nullable to use in `getRoute` function

navigator/runtime/src/main/java/com/freeletics/mad/navigator/internal/Navigate.kt
kboyarshinov: It's mostly for use with `NavBackStackEntry` and just `Bundle`, but yes, makes sense having separate `requireNotNull`

gabrielittner (approved):

PR #214, opened 2022-11-08
deploy docs directly from Github Actions

gabrielittner:
This removes the need for the `gh-pages` branch which will then be built and deployed.


kboyarshinov (approved):
:+1:

PR #215, opened 2022-11-08
new module for shared AndroidX navigation

gabrielittner:
Moves the AndroidX navigation code from the shared navigator module to a specific one. This will make it easier to experiment with a third implementation that doesn't have any AndroidX Navigation dependencies, by allowing us to completely remove AndroidX from the base module (in a follow up)


kboyarshinov (approved):
:+1:

PR #224, opened 2022-11-22
Add support for injecting dependencies into Composables

snappdevelopment:
This adds new functionality to inject classes into Composables, which have the `ComposeFragment` or `ComposeScreen` annotation applied.

Before this change users could provide UI specific dependencies via Composes' `LocalCompositionProvider`. This got removed in favour of injecting directly into the Composable.

Example:
We want to use an instance of `MyDependency` in the UI, which is provided through Dagger.
```
@ComposeFragment( //or @ComposeScreen
    scope = Example::class,
    parentScope = Singleton::class,
    stateMachine = ExampleStateMachine::class,
)
@Composable
internal fun ExampleUi(
    myDependency: MyDependency,
    state: ExampleState,
    sendAction: (ExampleAction) -> Unit,
) {
```

### DIFF
--- a/whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/WhetstoneCodeGenerator.kt
+++ b/whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/WhetstoneCodeGenerator.kt
@@ -220,10 +220,10 @@ public class WhetstoneCodeGenerator : CodeGenerator {
 
     private fun TopLevelFunctionReference.parameterTypes(): List<ClassName> {
         return parameters
-            .filter { it.fqName != null && it.name != "state" && it.name != "sendAction" }
+//            .filter { it.name != "state" && it.name != "sendAction" }
             .map {
 //                val typeFqName = it.type().asClassReference().fqName
-                ClassName(it.fqName!!.packageString(), it.fqName!!.shortName().asString())
+                ClassName(it.fqName?.packageString() ?: "FqNameIsNull", it.fqName?.shortName()?.asString() ?: "FqNameIsNull2")
             }
     }
 }
--- a/whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/WhetstoneCodeGenerator.kt
+++ b/whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/WhetstoneCodeGenerator.kt
@@ -1,6 +1,7 @@
 package com.freeletics.mad.whetstone
 
 import com.freeletics.mad.whetstone.codegen.FileGenerator
+import com.freeletics.mad.whetstone.codegen.util.AnvilCompilationExceptionTopLevelFunctionReference
 import com.freeletics.mad.whetstone.codegen.util.TopLevelFunctionReference
 import com.freeletics.mad.whetstone.codegen.util.composeFqName
 import com.freeletics.mad.whetstone.codegen.util.composeFragmentFqName
@@ -24,8 +25,11 @@ import com.squareup.anvil.compiler.api.GeneratedFile
 import com.squareup.anvil.compiler.api.createGeneratedFile
 import com.squareup.anvil.compiler.internal.fqNameOrNull
 import com.squareup.anvil.compiler.internal.reference.AnnotatedReference
+import com.squareup.anvil.compiler.internal.reference.AnvilCompilationExceptionAnnotationReference
+import com.squareup.anvil.compiler.internal.reference.AnvilCompilationExceptionParameterReference
 import com.squareup.anvil.compiler.internal.reference.ClassReference
 import com.squareup.anvil.compiler.internal.reference.classAndInnerClassReferences
+import com.squareup.anvil.compiler.internal.requireFqName
 import com.squareup.kotlinpoet.ClassName
 import java.io.File
 import org.jetbrains.kotlin.descriptors.ModuleDescriptor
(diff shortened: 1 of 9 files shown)


whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/WhetstoneCodeGenerator.kt
```diff
@@ -214,4 +223,25 @@ public class WhetstoneCodeGenerator : CodeGenerator {
     private fun FqName.packageString(): String {
         return pathSegments().joinToString(separator = ".")
     }
+
+    private fun TopLevelFunctionReference.parameterTypes(): List<ComposableParameter> {
+        return parameters
+            .filter { it.name != null && it.name != "state" && it.name != "sendAction" }
```
snappdevelopment, line 229: `state` and `sendAction` are specific to our codebase. Not sure how this could be prevented

gabrielittner: I think this is fine, for a first step. A better long term solution would be to see how we can get the type parameters of the state machine class and then match based on those. E.g. FooStateMachine which we have extends `StateMachine<FooState, FooAction>` so we can extract those 2 to use them here.

whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/WhetstoneCodeGenerator.kt
```diff
@@ -214,4 +223,25 @@ public class WhetstoneCodeGenerator : CodeGenerator {
...
+            .map {
+                val fqName = it.typeReference?.fqNameOrNull(module)
+                val packageString = fqName?.packageString()?.substringBeforeLast(".")
+                val shortName = fqName?.shortName()?.asString()
+                if(packageString == null || shortName == null) {
+                    throw AnvilCompilationExceptionTopLevelFunctionReference(
+                        functionReference = this,
+                        message = "Could not find class for ${it.name} in ${this.name}"
```
gabrielittner, line 237:
```suggestion
                        message = "Could not find class for parameter ${it.name} in ${this.name}"
```

whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/codegen/common/ComposeGenerator.kt
```diff
@@ -21,24 +20,30 @@ internal class ComposeGenerator(
 ) : Generator<ComposeData>() {
 
     internal fun generate(): FunSpec {
+        val parameterString = data.composableParameter
+            .joinToString { "${it.name} = ${it.className.propertyName}" }
+            .let { if(it.isNotBlank()) it.plus(", ") else it }
```
gabrielittner, line 25: I think it would make sense to build the whole composable call as a `CodeBlock` here and also make `state = currentState` and `sendAction = { ... }` named parameters

whetstone/compiler/src/test/kotlin/com/freeletics/mad/whetstone/codegen/FileGeneratorTestCompose.kt
```diff
@@ -23,6 +24,16 @@ internal class FileGeneratorTestCompose {
...
+        composableParameter = listOf(
+            ComposableParameter(
+                name = "testClass",
+                className = ClassName("com.test", "TestClass"),
+            ),
+            ComposableParameter(
+                name = "test",
+                className = ClassName("com.other", "TestClass2"),
+            )
+        )
```
gabrielittner, line 36: Can you move this to an extra test case? That adds the parameters by doing `data.copy(...)`

whetstone/compiler/src/test/kotlin/com/freeletics/mad/whetstone/codegen/FileGeneratorTestComposeFragment.kt
```diff
@@ -26,6 +27,16 @@ internal class FileGeneratorTestComposeFragment {
         fragmentBaseClass = fragment,
         navigation = null,
         navEntryData = null,
+        composableParameter = listOf(
```
gabrielittner, line 30: Same here

whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/WhetstoneCodeGenerator.kt
```diff
@@ -214,4 +223,25 @@ public class WhetstoneCodeGenerator : CodeGenerator {
...
         return pathSegments().joinToString(separator = ".")
     }
+
+    private fun TopLevelFunctionReference.parameterTypes(): List<ComposableParameter> {
+        return parameters
+            .filter { it.name != null && it.name != "state" && it.name != "sendAction" }
+            .map {
+                val fqName = it.typeReference?.fqNameOrNull(module)
+                val packageString = fqName?.packageString()?.substringBeforeLast(".")
+                val shortName = fqName?.shortName()?.asString()
```
gabrielittner, line 233: You could use `asClassNameOrNull` here and then just throw the exception for the null case https://github.com/square/anvil/blob/main/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/KotlinPoetUtils.kt#L41

whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/codegen/common/ComposeGenerator.kt
gabrielittner: Yes

whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/WhetstoneCodeGenerator.kt
snappdevelopment: This one is private. I could use the `asClassName` which is above it. However it throws it's own exception

snappdevelopment: I am using `asClassName` now. If `fqName` is null we throw our exception. If it is not null but the ClassName can't be found then `asClassName` throw its exception

gabrielittner: 👍

PR #226, opened 2022-11-22
Fix action lambda in compose code generation

snappdevelopment:
Fixes the lambda to use `it`. This was a mistake during refactoring


gabrielittner (approved):
We really need to add compilation tests

PR #233, opened 2022-12-07
move AndroidX navigation specific code more into androidx-nav module

gabrielittner:
#222 removed AndroidX navigation fror navigator-runtime. This is a follow up that moves some more AndroidX specific elements out of the runtime and prepares for making whetstone independent of AndroidX navigation.

Note: none of these changes affect the public API

- move methods that create destination ids from routes to the `androidx-nav` module
- to make the above change work the `NavigationResultRequest.Key` will now hold the route class instead of the id, the `AndroidXNavigationExecutor` will turn it into an id internally
-  added methods to whetstone to obtain a `SavedStateHandle`, `ViewModelStore` and `BaseRoute` for a given destination to hide AndroidX back stack related APIs
- generally replace `NavController` usages with `NavigationExecutor`
- add a composition local to obtain the `NavigationExecutor` (the NavController composition local will be removed when whetstone starts using the new APIs) and a find method for fragments

Whetstone changes will be an extra PR but can be seen here 575f956

### DIFF
--- a/navigator/runtime-fragment/src/main/java/com/freeletics/mad/navigator/fragment/FragmentExtensions.kt
+++ b/navigator/runtime-fragment/src/main/java/com/freeletics/mad/navigator/fragment/FragmentExtensions.kt
@@ -1,0 +1,16 @@
+package com.freeletics.mad.navigator.fragment
+
+import androidx.fragment.app.Fragment
+import androidx.navigation.fragment.findNavController
+import com.freeletics.mad.navigator.BaseRoute
+import com.freeletics.mad.navigator.internal.AndroidXNavigationExecutor
+import com.freeletics.mad.navigator.internal.InternalNavigatorApi
+import com.freeletics.mad.navigator.internal.NavigationExecutor
+import com.freeletics.mad.navigator.internal.requireRoute
+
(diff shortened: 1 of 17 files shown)


navigator/runtime-fragment/src/main/java/com/freeletics/mad/navigator/fragment/FragmentExtensions.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.freeletics.mad.navigator.BaseRoute
+import com.freeletics.mad.navigator.internal.AndroidXNavigationExecutor
+import com.freeletics.mad.navigator.internal.InternalNavigatorApi
+import com.freeletics.mad.navigator.internal.NavigationExecutor
+import com.freeletics.mad.navigator.internal.requireRoute
+
+public fun <T : BaseRoute> Fragment.requireRoute(): T = requireArguments().requireRoute()
+
+@InternalNavigatorApi
+public fun Fragment.findNavigationExectuor(): NavigationExecutor {
```
kboyarshinov, line 14:
```suggestion
public fun Fragment.findNavigationExecutor(): NavigationExecutor {
```

kboyarshinov (approved):
🚢

PR #234, opened 2022-12-07
use NavigationExecutor inside whetstone

gabrielittner:
Whetstone now has 0 direct dependencies on AndroidX Navigation, all it uses are our own navigation APIs, mainly through NavigationExecutor. This also means `LocalNavController` is being removed now.

Nice side effects
- we don't need the import alias workaround anymore
- the generated code got a bit simpler

### DIFF
--- a/whetstone/navigation/src/main/java/com/freeletics/mad/whetstone/internal/NavEntryViewModelProvider.kt
+++ b/whetstone/navigation/src/main/java/com/freeletics/mad/whetstone/internal/NavEntryViewModelProvider.kt
@@ -7,34 +7,35 @@ import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.createSavedStateHandle
 import androidx.lifecycle.viewmodel.initializer
 import androidx.lifecycle.viewmodel.viewModelFactory
-import androidx.navigation.NavBackStackEntry
 import com.freeletics.mad.navigator.BaseRoute
+import com.freeletics.mad.navigator.internal.NavigationExecutor
 import kotlin.reflect.KClass
 
 /**
- * Creates a [ViewModel] for the given [entry]. The `ViewModel.Factory` will use [scope] to lookup
- * a parent component instance. That component will then be passed to the given [factory] together
- * with a [SavedStateHandle] and the passed in [route].
+ * Creates a [ViewModel] for the given [destination]. The `ViewModel.Factory` will use [parentScope]
+ * to lookup a parent component instance. That component will then be passed to the given [factory]
+ * together with a [SavedStateHandle] and the passed in [destination].
  *
  * To be used in generated code.
  */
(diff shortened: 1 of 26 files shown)


whetstone/navigation/src/main/java/com/freeletics/mad/whetstone/internal/NavEntryViewModelProvider.kt
```diff
@@ -7,34 +7,35 @@ import androidx.lifecycle.ViewModelProvider
...
+ * to lookup a parent component instance. That component will then be passed to the given [factory]
+ * together with a [SavedStateHandle] and the passed in [destination].
  *
  * To be used in generated code.
  */
 @InternalWhetstoneApi
-public inline fun <reified T : ViewModel, D : Any, R : BaseRoute> viewModel(
-    entry: NavBackStackEntry,
+public inline fun <reified T : ViewModel, D : Any, R : BaseRoute> navEntryViewModel(
+    destination: KClass<R>,
```
gabrielittner, line 23: We use this for looking up both the route (doesn't need to happen in generated code) and the viewModelStore (removes the need for `NavBackStackEntry`)

whetstone/navigation/src/main/java/com/freeletics/mad/whetstone/internal/NavEntryViewModelProvider.kt
```diff
@@ -7,34 +7,35 @@ import androidx.lifecycle.ViewModelProvider
...
+    executor: NavigationExecutor,
```
gabrielittner, line 24: We use this for looking up both the route (doesn't need to happen in generated code) and the viewModelStore (removes the need for `NavBackStackEntry`)

whetstone/navigation/src/main/java/com/freeletics/mad/whetstone/internal/NavEntryViewModelProvider.kt
```diff
@@ -7,34 +7,35 @@ import androidx.lifecycle.ViewModelProvider
...
     context: Context,
-    scope: KClass<*>,
+    parentScope: KClass<*>,
```
gabrielittner, line 26: Makes it a bit clearer than before

whetstone/navigation/src/main/java/com/freeletics/mad/whetstone/internal/NavEntryViewModelProvider.kt
```diff
@@ -7,34 +7,35 @@ import androidx.lifecycle.ViewModelProvider
...
     destinationScope: KClass<*>,
-    route: R,
-    noinline findEntry: (Int) -> NavBackStackEntry,
```
gabrielittner, line 28: Is already part of NavigationExecutor

PR #237, opened 2022-12-09
replace PermissionResult enum with a sealed class

gabrielittner:
With this the result matches what the platform provides and does not try to interpret the values (in a wrong way).

### DIFF
--- a/navigator/runtime/src/main/java/com/freeletics/mad/navigator/ResultOwner.kt
+++ b/navigator/runtime/src/main/java/com/freeletics/mad/navigator/ResultOwner.kt
@@ -1,17 +1,14 @@
 package com.freeletics.mad.navigator
 
-import android.app.Activity
+import android.content.Context
 import android.os.Parcel
 import android.os.Parcelable
 import androidx.activity.result.contract.ActivityResultContract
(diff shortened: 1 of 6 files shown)


navigator/runtime/src/main/java/com/freeletics/mad/navigator/ResultOwner.kt
```diff
@@ -74,31 +71,51 @@ public class ActivityResultRequest<I, O> internal constructor(
...
+         *
+         * **Reasons:**
+         * - the user denied the request
+         *     - `shouldShowRationale` is `true`
+         * - Android 10 and below: the user selected to not be asked again
+         *     - `shouldShowRationale` is `false`
+         * - Android 11+: the user denied the request twice already and the system won't ask again
+         *     - `shouldShowRationale` is `false`
+         * - Android 11+: the user dismissed the notification request without making a choice
+         *     - `shouldShowRationale` is `false` if the user dit not deny the permission before
```
josirichter, line 109:
```suggestion
         *     - `shouldShowRationale` is `false` if the user did not deny the permission before
```

josirichter (approved):

PR #240, opened 2022-12-12
share collecting nav events

gabrielittner:
Adds `collectAndHandleNavEvents` to the shared runtime which will deal with collecting the nav events flow. I'm moving more code to `runtime` were possible to make adding a third runtime implementation easier.

### DIFF
--- a/navigator/runtime/src/main/java/com/freeletics/mad/navigator/internal/Navigate.kt
+++ b/navigator/runtime/src/main/java/com/freeletics/mad/navigator/internal/Navigate.kt
@@ -1,35 +1,51 @@
 package com.freeletics.mad.navigator.internal
 
+import android.annotation.SuppressLint
 import androidx.activity.result.ActivityResultLauncher
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.flowWithLifecycle
 import com.freeletics.mad.navigator.ActivityResultRequest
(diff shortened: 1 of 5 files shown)


navigator/runtime/src/main/java/com/freeletics/mad/navigator/internal/Navigate.kt
```diff
@@ -50,7 +66,7 @@ public fun navigate(
             (launcher as ActivityResultLauncher<Any?>).launch(event.permissions)
         }
         is NavEvent.DestinationResultEvent<*> -> {
-            controller.deliverResult(event.key, event.result)
+            savedStateHandleFor(event.key.route)[event.key.requestKey] = event.result
```
kboyarshinov, line 69: What about still using `deliverResult` to not duplicate map access? or do you have other plans for it?
```suggestion
            deliverResult(event.key, event.result)
```

kboyarshinov (approved):
👍

navigator/runtime/src/main/java/com/freeletics/mad/navigator/internal/Navigate.kt
gabrielittner: I do plans to make the contents of key internal and move this and some other result based logic here, but that was meant for another pr

PR #241, opened 2022-12-12
clean up public api, start using poko, add few more tests

gabrielittner:
- make subclasses of `NavDestination` internal but annotate them with `@PublishedApi`, this will hide them from consumers but still makes them public in bytecode so that they can be used by the reified inline functions that we use to build them
- same for `NavEventNavigator` apis that have reified overlaods
- now that all `KClass<Route>` usages are internal, create a new internal `DestinationId` value/inline class to wrap it
- use [Poko](https://github.com/drewhamilton/Poko) instead of data classes, it generates equals/hashCode/toString but no copy or component methods which we don't need/want and which risk unintentionally making binary incompatible changes
- add tests for `ResultOwner`

(sorry this got a bit big in scope, but there are no logic/behavior changes here)


navigator/runtime/src/test/java/com/freeletics/mad/navigator/NavEventNavigatorTest.kt
```diff
@@ -232,7 +229,7 @@ public class NavEventNavigatorTest {
     public fun `registerForPermissionsResult after read is disallowed`(): Unit = runBlocking {
         val navigator = TestNavigator()
 
-        navigator.permissionsResultRequests
+        navigator.activityResultRequests
```
kboyarshinov, line 232: should this have been changed?

kboyarshinov (approved):
Nice 👍 LGTM

navigator/runtime/src/test/java/com/freeletics/mad/navigator/NavEventNavigatorTest.kt
gabrielittner: Not in this pr 😅 
```suggestion
        navigator. permissionsResultRequests
```

PR #242, opened 2022-12-13
share collecting navigation results, make more of NavigationResultRequest.Key internal

gabrielittner:
Similar to the shared `collectAndHandleNavEvents`, the main runtime now has `collectAndHandleNavigationResults` which will take care of collecting and forwarding navigation results. `deliverResult` was also removed from the executor and instead is done in the nav event code since we are exposing the saved state handle anyways. The latter allows us to make the contents of the Key completely internal.


kboyarshinov (approved):
:+1:

PR #243, opened 2022-12-13
don't allow passing NavRoot to navigateBackTo, instead add clear option to navigateToRoot

gabrielittner:
This will make implementing custom back stacks easier, because it ensures that the pop up to stays within the current back stack which starts with a `NavRoot`. In practice this means that if you want to go to the root of your stack or even the default back stack you need to use `navigateToRoot`. That method is now exposing `saveCurrentRootState` which previously was always true internally and allows dropping the state of the current stack.

With this the usages of navigateToRoot are
- bottom nav selected: `navigateToRoot(root, restoreRootState = true, saveCurrentRootState  = true)`
- current bottom nav re-selected: `navigateToRoot(root, restoreRootState = false, saveCurrentRootState  = false)` (we have true here right now but I guess it doesn't do anything because we are going to the same root and not restoring anything)
- navigating to another tab in code:  `navigateToRoot(root, restoreRootState = false, saveCurrentRootState  = true)`
- going back to start:  `navigateToRoot(root, restoreRootState = false, saveCurrentRootState  = false)`

### DIFF
--- a/navigator/androidx-nav/src/main/java/com/freeletics/mad/navigator/internal/AndroidXNavigationExecutor.kt
+++ b/navigator/androidx-nav/src/main/java/com/freeletics/mad/navigator/internal/AndroidXNavigationExecutor.kt
@@ -25,10 +25,14 @@ public class AndroidXNavigationExecutor(
         controller.navigate(route.destinationId(), route.getArguments())
     }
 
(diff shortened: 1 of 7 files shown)


kboyarshinov (approved):
👍

PR #244, opened 2022-12-14
make ActivityRoute Parcelable

gabrielittner:
Currently it is not `Parcelable` because putting it into the `Intent` would crash any third party app launched by it. So consumers can make it Parcelable for internal use and then we will add it to the extras. The issue is that this makes deep links a lot more complex because we can't pass through the route and then handle it. With this `ActivityRoute` becomes `Parcelable` and the default implementation of `fillInIntent` will add it to the extras. Other places like the activity navigator don't need to deal with it anymore. When implementing route for external destinations users can override `fillInIntent` to prevent the route from being added and therefor avoid the crash.

### DIFF
--- a/navigator/runtime/src/main/java/com/freeletics/mad/navigator/NavRoute.kt
+++ b/navigator/runtime/src/main/java/com/freeletics/mad/navigator/NavRoute.kt
@@ -28,15 +28,15 @@ public interface NavRoot : BaseRoute
 /**
  * Represents the route to a destination.
  *
- * When the implementing class is [Parcelable], the instance of route will be put into the
- * navigation arguments and is then available to the target screens. Do not do this in case this
- * route leads to a different app.
+ * By default the instance of this class will be added to th resulting `Intent` and can be accessed
+ * in the launched `Activity` by calling [getRoute] or [requireRoute].
+ *
+ * If this route leads to another app it is required to override [fillInIntent] and at least return
+ * an empty [Intent] from it.
  */
-public interface ActivityRoute {
-    public fun fillInIntent(): Intent = EMPTY_INTENT
-
-    public companion object {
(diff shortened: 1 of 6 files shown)


navigator/runtime/src/main/java/com/freeletics/mad/navigator/NavRoute.kt
```diff
@@ -28,15 +28,15 @@ public interface NavRoot : BaseRoute
 /**
  * Represents the route to a destination.
  *
- * When the implementing class is [Parcelable], the instance of route will be put into the
- * navigation arguments and is then available to the target screens. Do not do this in case this
- * route leads to a different app.
+ * By default the instance of this class will be added to th resulting `Intent` and can be accessed
```
kboyarshinov, line 31:
```suggestion
 * By default the instance of this class will be added to the resulting `Intent` and can be accessed
```

kboyarshinov (approved):
LGTM :+1: 

what do you think of adding something like `ExternalActivityRoute` to facilitate corrent external activity use case?

```kotlin
public abstract class ExternalActivityRoute : ActivityRoute {
    final override fun fillInIntent(): Intent {
        return externalIntent()
    }
    
    public abstract fun externalIntent(): Intent
}
```

gabrielittner:
> what do you think of adding something like ExternalActivityRoute to facilitate corrent external activity use case?

Added it but in a slightly different way. I've made `ActivityRoute` sealed and then also added `InternalActivityRoute`. This one has a final override of fillInIntent which means it is guaranteed that these have the route extra. `ExternalActivityRoute` then has the same default implementation with an empty intent that we had on the base interface before.

gabrielittner:
snapshot publishing failed because of a 502 from sonatype

PR #245, opened 2022-12-14
share handling activity results, merge activity and permission handling

gabrielittner:
Instead if treating permission requests and activity results completely separately this introduces an intermediate `ContractResultOwner` interface for both of their result owners to allow us to handle them in the same way. Both use activity result contracts, the only difference is that the permission results get processed a bit before being passed on. With this change the whole registration and waiting for results part is merged internally. The public API stays separate.


snappdevelopment (approved):
Looks good!

PR #246, opened 2022-12-14
stop generating ViewModels

gabrielittner:
We now have a `StoreViewModel` in the runtime that can hold on to arbitrary objects to make them survive configuration changes. There is also `CloseableComponent` base interface. In it's `onCleared` method `StoreViewModel` will automatically close all closeables in objects that it holds that also implement that interface.

(these 2 new types are duplicated, one in the navigation runtime and one in whetstone for the version without navigation) 

**Navigation**

On the navigation side the `viewModelStoreFor` method was replaced with a `storeFor` method that will return an instance of the `StoreViewModel`. The `StoreViewModel` will be tied to the back stack entry and cleared when the entry is removed from the back stack.

**Whetstone**

- stop generating view models
- all generated components will implement `CloseableComponent` instead of generating their own closeable set (nav entry components still need to override it to add a qualifier)
- the runtime methods to obtain a view model are changed to obtain a component instead
	- on the navigation side `storeFor` is used to obtain a store  and then get or create the component
	- on the other side the whetstone StoreViewModel is obtained using the hosts (fragments/composables)`ViewModelStore` 

`NavEntryComponentGetter` could be renamed to `NavEntryComponentFactory` now but I didn't want to add more things to this pr.


whetstone/runtime/src/main/java/com/freeletics/mad/whetstone/internal/CloseableComponent.kt
```diff
@@ -0,0 +1,7 @@
+package com.freeletics.mad.whetstone.internal
+
+import java.io.Closeable
+
+public interface CloseableComponent {
```
kboyarshinov, line 5: What do you think about making Component itself `Closable`, i.e.
```
public interface WhetstoneComponent : Closeable {
    public val closables: Set<Closeable>

    override fun close() {
        closables.forEach { it.close() }
    }
}
```

Since it's generated, it's unlikely to be overriden anyway. This way component itself can be directly added to `ViewModel`'s internal Closables and `CloseableComponent` won't be necessary.

kboyarshinov (approved):
:shipit:

PR #258, opened 2022-12-16
fix(navigator-runtime): write `NavigationResultRequest.Key.requestKey` to Parcel

hoc081098:


gabrielittner (approved):
Nice catch, thanks for fixing this

PR #268, opened 2023-01-02
automate more parts of the publishing process

gabrielittner:
With this it is not neccessary to push commits to `main` anymore to do a release and as much as possible is automated. The remaining steps are:
1. update changelog (through a pr), we could probably also have some automation here but I personally don't like these auto generated changelogs
2. create and push a tag for the new version

To make this possible there are a few changes:
- slightly updated the `CHANGELOG.md` format
- the release workflow will extract the section for the new release and automatically create a Github release with it
- `VERSION_NAME` is not comitted anymore and instead always computed
- for releases we tag the tag name as version
- for snapshots we take the last tag increment the last number by 1 and then append SNAPSHOT
- the readme contains `<latest-version>` pacehoders and one of those badges that shows the last version
- for the website we automatically replace `<latest-version>` with the last tag and remove the badge
- snapshot and website workflows run after the release so that there is a snapshot for the next release and the website contains the latest version

The only part that I don't like that much is the `<latest-version>`  placeholder in the README, I prefer when it just has the full declaration ready to copy, but at least it's like this on the website.

Already tested this [here](http://github.com/gabrielittner/renderer)


CHANGELOG.md
```diff
@@ -73,19 +87,18 @@ to the annotation.
 
 ### Other
 
-- Migrate to Gradle version catalog
+- Migrate to Gradle ## catalog
```
kboyarshinov, line 90: `version` is missing

```suggestion
- Migrate to Gradle version catalog
```

kboyarshinov (approved):
Nice :+1: :shipit:

CHANGELOG.md
gabrielittner: That was the search + replace 😅

.github/workflows/publish-release.yml
```diff
@@ -23,6 +24,16 @@ jobs:
       - name: Install Android SDK
         run: ./.github/android-sdk.sh
 
+      - name: Get release notes
+        run: |
+          echo "RELEASE_NOTES<<EOF" >> $GITHUB_ENV
+          echo "$(awk '/^## ${{ github.ref_name }}/{flag=1;next}/^## /{flag=0}flag' CHANGELOG.md)" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
gabrielittner: Got it from stack overflow :D

PR #276, opened 2023-01-13
fix deep links that contain ActivityRoutes

gabrielittner:
The code was only expecting `BaseRoute` but it can bei either `ActivityRoute` or `BaseRoute`.


kboyarshinov (approved):
:shipit:

PR #277, opened 2023-01-13
update to Anvil 2.4.4 and use new TopLevelFunctionReference

### LINKED ISSUE #256: Support generics for objects injected into composables
This would be a follow up to #224 which generally supports adding more parameters to a `@ComposeScreen` or `@ComposeFragment` composable to get an instance of them injected. This initial support does not handle generics on the injected parameters, so it is for example not possible to inject a `Set<Foo>`

gabrielittner:
Updates Kotlin to 1.8.0, Compose Compiler to 1.4.0 and Anvil to 2.4.4. Anvil now offers a `TopLevelFunctionReference` abstraction which allows us to delete our own copy and improves parameter parsing over what we are currently doing. The latter automatically gives us support for objects with type parameters injecting into composable functions. Our tests that run Anvil are also using Anvil's helper for the set up now, I think the issue before was in incompatibility with the kotlin-compile-testing version wer were using.

Closes #256


whetstone/compiler-test/src/test/kotlin/com/freeletics/mad/whetstone/codegen/TestHelper.kt
```diff
@@ -48,26 +49,30 @@ private fun compile(fileName: String, source: String, output: String) {
...
 }
 
+@OptIn(ExperimentalCompilerApi::class, ExperimentalAnvilApi::class)
 private fun compileWithAnvil(fileName: String, source: String, output: String) {
-    val compilation = KotlinCompilation().apply {
-        configure()
+    val compilation = AnvilCompilation().apply {
         configureAnvil()
+        kotlinCompilation.configure()
+        useIR(true)
```
gabrielittner, line 57: This is a helper from Anvil which was previously not working for us. It has its own `configureAnvil` method so that we don't need our own below anymore

whetstone/compiler-test/src/test/kotlin/com/freeletics/mad/whetstone/codegen/FileGeneratorTestCompose.kt
```diff
@@ -576,7 +594,11 @@ internal class FileGeneratorTestCompose {
 
               public val testClass: TestClass
 
-              public val testClass2: TestClass2
+              public val test: TestClass2
```
gabrielittner, line 597: This changed because we are now using the parameter name here instead of a name derived from the class.

snappdevelopment (approved):
Nice improvement 👍 
I left one optional comment below

PR #280, opened 2023-01-13
remove NONE destination case from codegen and tests

gabrielittner:
The `NONE` enum value was removed a long time ago, so we don't actually need to support it in the codegen. This primarily means that we don't need navigation and navigation + destination tests anymore.


kboyarshinov (approved):
👍

PR #281, opened 2023-01-13
run Anvil in all codegen tests

### LINKED ISSUE #249: Run whetstone tests with actual source files
Currently we are testing the codegen by using our abstracted models as input to see that the output matches what we expect. This means that the tests don't cover the conversion from getting `KtFile` instances to our model

gabrielittner:
Builds on top of #280 so that there are less tests to update. With this each test defines a source file that matches the data that we define for that tests. This source file replaces the fixture classes we had before and we will do an additional check by running a Kotlin compilation with Anvil over it and also check the output of that. We can probably remove the manual `FileGenerator` invocation and data objects as a follow up but not 100% sure yet. 

The renderer tests have a few more changes because the `baseName` there was `Test` and not `TestRenderer` but the renderer was stilled called `TestRenderer`. 

Closes #249

### DIFF
--- a/whetstone/compiler-test/src/test/kotlin/com/freeletics/mad/whetstone/codegen/TestHelper.kt
+++ b/whetstone/compiler-test/src/test/kotlin/com/freeletics/mad/whetstone/codegen/TestHelper.kt
@@ -5,39 +5,117 @@ import com.freeletics.mad.whetstone.ComposeFragmentData
 import com.freeletics.mad.whetstone.ComposeScreenData
 import com.freeletics.mad.whetstone.RendererFragmentData
 import com.google.common.truth.Truth.assertThat
+import com.squareup.anvil.compiler.AnvilCommandLineProcessor
+import com.squareup.anvil.compiler.AnvilComponentRegistrar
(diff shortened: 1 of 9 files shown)


whetstone/compiler-test/src/test/kotlin/com/freeletics/mad/whetstone/codegen/TestHelper.kt
```diff
@@ -5,39 +5,118 @@ import com.freeletics.mad.whetstone.ComposeFragmentData
...
+        PluginOption(
+            pluginId = anvilCommandLineProcessor.pluginId,
+            optionName = "generate-dagger-factories-only",
+            optionValue = "false"
+        ),
+        PluginOption(
+            pluginId = anvilCommandLineProcessor.pluginId,
+            optionName = "disable-component-merging",
+            optionValue = "false"
         )
```
gabrielittner, line 120: Anvil has a helper for this in the utils test fixture but I got some weird compilation errors with that. I'll try again after the Kotlin 1.8/Anvil 2.4.4 update.

kboyarshinov (approved):
LGTM 👍

PR #292, opened 2023-02-06
move NavEvent to internal package

gabrielittner:
It is already marked as internal so it can also go into that package.


kboyarshinov (approved):
:+1:

PR #295, opened 2023-02-08
add AppScope and make it the default parent/destinationScope

### LINKED ISSUE #284: Add an `AppScope` marker class to the library
We can provide this and have it as the default value for both `destinationScope` and `parentScope`. That way the standard case only needs to pass the `route` and the `stateMachine` as parameters

gabrielittner:
This makes it unnecessary to explicitly specify `parentScope` and `destinationScope` for most cases. In our app you basically only need parent scope you have a nav entry component that you use as parent. The reason why I went with `AppScope` over using `Singleton` like we currently do is that it is what Anvil's docs suggest and also what is used by others. Also makes it so that all scoping will happen through `@ScopeTo(...)` and not a mix of annotations.

closes #284

### DIFF
--- a/docs/whetstone/get-started.md
+++ b/docs/whetstone/get-started.md
@@ -58,7 +58,7 @@ it falls into the `Fragment` category.
     ```kotlin
     @ComposeScreen(
         scope = ExampleScope::class,
-        parentScope = AppScope::class,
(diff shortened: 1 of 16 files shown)


kboyarshinov (approved):
LGTM :shipit:

PR #298, opened 2023-02-09
fix name shadowed warning and enforce no warnings in tests

gabrielittner:
Tests now use allWarningsAsError when compiling the generated code to avoid this from happening again.


kboyarshinov (approved):
👍

PR #300, opened 2023-02-11
don't use lifecycle in compose navigation setup

gabrielittner:
See coments on the code

### DIFF
--- a/navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavigationSetup.kt
+++ b/navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavigationSetup.kt
@@ -10,7 +10,6 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.ui.platform.LocalContext
-import androidx.compose.ui.platform.LocalLifecycleOwner
 import com.freeletics.mad.navigator.ContractResultOwner
 import com.freeletics.mad.navigator.NavEventNavigator
 import com.freeletics.mad.navigator.internal.collectAndHandleNavEvents
@@ -24,7 +23,6 @@ import com.freeletics.mad.navigator.internal.deliverResult
 @Composable
 public fun NavigationSetup(navigator: NavEventNavigator) {
     val executor = LocalNavigationExecutor.current
-    val lifecycleOwner = LocalLifecycleOwner.current
     val context = LocalContext.current
 
     val activityLaunchers = navigator.activityResultRequests.associateWith {
@@ -38,16 +36,16 @@ public fun NavigationSetup(navigator: NavEventNavigator) {
     }
 
     val backDispatcher = LocalOnBackPressedDispatcherOwner.current!!.onBackPressedDispatcher
-    DisposableEffect(lifecycleOwner, backDispatcher, navigator) {
-        backDispatcher.addCallback(lifecycleOwner, navigator.onBackPressedCallback)
+    DisposableEffect(backDispatcher, navigator) {
+        backDispatcher.addCallback(navigator.onBackPressedCallback)
 
         onDispose {
             navigator.onBackPressedCallback.remove()
         }
     }
 
-    LaunchedEffect(lifecycleOwner, executor, navigator) {
-        navigator.collectAndHandleNavEvents(lifecycleOwner.lifecycle, executor, activityLaunchers)
+    LaunchedEffect(executor, navigator) {
+        navigator.collectAndHandleNavEvents(executor, activityLaunchers)
     }
 }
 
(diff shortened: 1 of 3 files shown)


navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavigationSetup.kt
```diff
@@ -38,16 +36,16 @@ public fun NavigationSetup(navigator: NavEventNavigator) {
...
 
     val backDispatcher = LocalOnBackPressedDispatcherOwner.current!!.onBackPressedDispatcher
-    DisposableEffect(lifecycleOwner, backDispatcher, navigator) {
-        backDispatcher.addCallback(lifecycleOwner, navigator.onBackPressedCallback)
+    DisposableEffect(backDispatcher, navigator) {
+        backDispatcher.addCallback(navigator.onBackPressedCallback)
 
         onDispose {
             navigator.onBackPressedCallback.remove()
         }
```
gabrielittner, line 44: We already call `remove` in `onDispose` so there is no need to pass a lifecycle to `addCallback` which would achieve the same thing.

navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavigationSetup.kt
```diff
@@ -38,16 +36,16 @@ public fun NavigationSetup(navigator: NavEventNavigator) {
...
     }
 
-    LaunchedEffect(lifecycleOwner, executor, navigator) {
-        navigator.collectAndHandleNavEvents(lifecycleOwner.lifecycle, executor, activityLaunchers)
+    LaunchedEffect(executor, navigator) {
+        navigator.collectAndHandleNavEvents(executor, activityLaunchers)
```
gabrielittner, line 48: The LaunchedEffect gets cancelled automatically if the composition leaves the screen so no need for a lifecycle here.

navigator/runtime-fragment/src/main/java/com/freeletics/mad/navigator/fragment/HandleNavigation.kt
```diff
@@ -36,7 +38,9 @@ public fun handleNavigation(fragment: Fragment, navigator: NavEventNavigator) {
     dispatcher.addCallback(fragment, navigator.onBackPressedCallback)
 
     lifecycle.coroutineScope.launch {
-        navigator.collectAndHandleNavEvents(lifecycle, executor, activityLaunchers)
+        lifecycle.repeatOnLifecycle(Lifecycle.State.RESUMED) {
+            navigator.collectAndHandleNavEvents(executor, activityLaunchers)
+        }
```
gabrielittner, line 43: `repeatOnLifecycle` is what `flowOnLifecycle` does internally, so for `Fragment` nothing changes.

navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavigationSetup.kt
hoc081098: This means that the events can be handled even if the Composable is not resumed. Is this safe?
Ref link: https://manuelvivo.dev/consuming-flows-compose
What do you think about that 😂?

gabrielittner: I think that is valid for things where collecting the flows causes work to happen. For example if that flow computes something, emits based on a timer or observes a database. In this case we are just receiving events from a channel and if something would be emitted while the app is not resumed the issue would be on the side what is triggering navigation events in your app while it is not resumed.

kboyarshinov (approved):
:+1:

navigator/runtime-compose/src/main/java/com/freeletics/mad/navigator/compose/NavigationSetup.kt
hoc081098: > I think that is valid for things where collecting the flows causes work to happen. For example if that flow computes something, emits based on a timer or observes a database. In this case we are just receiving events from a channel and if something would be emitted while the app is not resumed the issue would be on the side what is triggering navigation events in your app while it is not resumed.

If we use `NavEventNavigator` in a ViewModel to trigger the navigation, that is weird to check lifecycle state before triggering in the navigation.

I think we can trigger the navigation without lifecycle problem, and let consumer side observe nav events flow only when app is in resumed state.
How about 
```kotlin
val owner = LocalLifecycleOnwer.current
LaunchedEffect(owner, ...) {
    owner.repeatOnLifecycle(RESUMED) {
        collectAndHandle...
    }
}
```

gabrielittner: What is triggering the call to `NavEventNavigator` in the ViewModel in this case? My assumption is that user actions going through the ViewModel wouldn't happen in this state because the user shouldn't be able to interact with something that isn't resumed. If it is some work that is going on inside the ViewModel, shouldn't that also be stopped?

Having said that I will revert the event specific parts of this pr, it makes sense to keep it even if in our Androidx free implementation this is based on the activity lifecycle.

gabrielittner: Reverted here d02b1302. Thanks for kicking of this coversation

PR #301, opened 2023-02-11
use DisposeOnViewTreeLifecycleDestroyed for compose in fragments

gabrielittner:
Notiiced that the docs now recommend `DisposeOnViewTreeLifecycleDestroyed` instead of `DisposeOnLifecycleDestroyed`https://developer.android.com/jetpack/compose/interop/interop-apis#compose-in-fragments


kboyarshinov (approved):
:+1:

PR #302, opened 2023-02-12
add navigator:runtime-experimental

gabrielittner:
This creates a new navigator runtime module that is meant to be the place to experiment with an AndroidX navigation free implementation of our navigation APIs. The module is meant as a drop in replacement for the existing compose runtime, meaning you should just be able to add the following snippet to settings.gradle and everything will work without further changes:
```
gradle.beforeProject {
    configurations.all {
        resolutionStrategy.eachDependency {
            if (requested.group == "com.freeletics.mad" && requested.name == "navigator-compose") {
                useTarget("com.freeletics.mad:navigator-experimental:${requested.version}")
            }
        }
    }
}
```

Added files:
- `NavDestination` is a 100% copy
- `NavHost` is a copy with the same public signature but the implementation was replaced with TODOs
- `NavigationSetup` is a 100% copy
- added stub implementation of `NavigationExecutor`

Additionally for both the `NavDestination` subclasses have been unified for easier access.

CI will be fine after #300 is merged.


navigator/runtime-compose/api/navigator-runtime-compose.api
```diff
@@ -21,7 +25,9 @@ public final class com/freeletics/mad/navigator/compose/NavigationSetupKt {
 	public static final fun NavigationSetup (Lcom/freeletics/mad/navigator/NavEventNavigator;Landroidx/compose/runtime/Composer;I)V
 }
 
-public final class com/freeletics/mad/navigator/compose/ScreenDestination : com/freeletics/mad/navigator/compose/NavDestination {
+public final class com/freeletics/mad/navigator/compose/ScreenDestination : com/freeletics/mad/navigator/compose/ContentDestination {
```
kboyarshinov, line 28: These destinations by being public expose internal `ContentDestination`

navigator/runtime-experimental/api/runtime-experimental.api
```diff
@@ -0,0 +1,40 @@
+public final class com/freeletics/mad/navigator/compose/ActivityDestination : com/freeletics/mad/navigator/compose/NavDestination {
+	public synthetic fun <init> (Lkotlin/reflect/KClass;Landroid/content/Intent;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
+}
+
+public final class com/freeletics/mad/navigator/compose/BottomSheetDestination : com/freeletics/mad/navigator/compose/ContentDestination {
```
kboyarshinov, line 5: same here

kboyarshinov (approved):
LGTM 👍

navigator/runtime-compose/api/navigator-runtime-compose.api
gabrielittner: Yes, the reason why this works and is ok is that these subclasses are also internal but use `@PublishedApi`, so from a binary perspective this is public but from a source perspective all of this is internal.

PR #304, opened 2023-02-12
move NavigationExecutor.Store to standalone class in runtime

gabrielittner:
Extracted this from my experimental navigator implementation. This way all navigator implementations can share the same store. For this to work the store itself isn't a `ViewModel` anymore (the experimental implementation tries to just have one ViewModel per Activity instead of one per back stack entry), so the AndroidX implementation still has a `StoreViewModel` that is just delegating to the actual store. The only thing we are loosing is the `addCloseable` which needed to be replaced with manually iterating. 

Also added tests for the store.


gabrielittner:
I haven't fully made up my mind about that yet. It would introduce a navigator dependency in the basic whetstone artifacts, but maybe this is fine.

PR #312, opened 2023-02-20
automatically find ViewRenderer.Factory instead of requiring a parameter

### LINKED ISSUE #257: Remove rendererFactory parameter from RendererFragment annotation
We expect the factory to be a nested class anyways so we can just find it automatically inside the code generator. Either by name or (better) by type.

gabrielittner:
This is done by going through all nested classes of an annotated renderer and then finding one that extends `ViewRenderer.Factory`.

Closes #257


kboyarshinov (approved):
LGTM 👍

PR #313, opened 2023-02-20
make state and sendAction optional and prepare to make them more dynamic

### LINKED ISSUE #310: Make compose parameters optional
Right now each annotated composable is forced to have a `state` parameter and a `sendAction` parameter. This sometimes leads to unused parameter warnings in simple screens that only show static information or have no interactions. Recognize the parameters not being present and then adapt the generated code to not pass on these parameters.

gabrielittner:
Closes #310.

This does 2 things
- we are searching for the `state` and `sendAction` parameter in the annotated composable and put them into our models, this logic allows for them to not be present (matching is still done by name)
- the code generation will use the parameters values and also skip generating the relevant bits if the parameters are not present, the state machine will still be collected in any case as it might do some work even if there is no communication with the UI


kboyarshinov (approved):
LGTM 👍

PR #314, opened 2023-02-21
make compose parameters more dynamic by not forcing names

gabrielittner:
Instead of expecting `state` and `sendAction` to have specific names the parsing code will not dynamically find them based on the type parameters of `StateMachine`.

### DIFF
--- a/whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/parser/Reference.kt
+++ b/whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/parser/Reference.kt
@@ -5,14 +5,20 @@ import com.squareup.anvil.annotations.ExperimentalAnvilApi
 import com.squareup.anvil.compiler.internal.reference.AnnotatedReference
 import com.squareup.anvil.compiler.internal.reference.AnnotationReference
 import com.squareup.anvil.compiler.internal.reference.AnvilCompilationExceptionAnnotationReference
+import com.squareup.anvil.compiler.internal.reference.AnvilCompilationExceptionClassReference
 import com.squareup.anvil.compiler.internal.reference.ClassReference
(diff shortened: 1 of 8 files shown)


whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/parser/Reference.kt
```diff
@@ -75,3 +91,50 @@ internal val ClassReference.packageName: String
...
+    // e.g. for a StateMachine and State this would return 0
+    val index = superType.asClassReference().typeParameters.indexOfFirst { it.name == parameter }
+    // this is the type that is used in the implementation
+    // e.g. for ... : StateMachine<S, A> this would be S
+    val unwrappedType = superType.unwrappedTypes[index]
+    // resolve the type using the implementation class
+    val resolved = unwrappedType.resolveGenericTypeOrNull(this)
+    if (resolved != null) {
+        return resolved.asTypeName()
+    }
```
kboyarshinov, line 114: 🤯

kboyarshinov (approved):
LGTM 👍

whetstone/compiler/src/main/kotlin/com/freeletics/mad/whetstone/parser/Reference.kt
gabrielittner: This is already the simple version 😆 Initially I had `superTypeReference` return a list of type references for the whole hierarchy between the given class and the state machine interface and then this method would walk through the hierarchy and keep track of how the name of the type parameters change from one hierarchy level to another

kboyarshinov: Nice! Well done! :shipit:

PR #315, opened 2023-02-21
change startRoute to alwasy be a NavRoot, fix destinationCallback type

gabrielittner:
- For simplicity the API forces the start destination to use `NavRoot` which allows to make the assumption that each back stack always starts with a `NavRoot`
- Fixes the `destinationChangedCallback` to receive a `BaseRoute`, this was actually a bug until now which we just didn't notice because we didn't use `NavRoot` in compose navigation. If we would have this would crash.


kboyarshinov (approved):
👍

PR #322, opened 2023-03-13
fix type parameter parsing

gabrielittner:
Found an issue with the type resolving while preparing to update our internal code base. I'm not entirely sure why it doesn't work but it has something to do with mixing classes from different sources. The new test case would not fail if `TestStateMachine` is defined in the source string in the test instead of as regular class in the test source set and it would also not fail if `TestStateMachine` would extend `StateMachine` instead of `FlowReduxStateMachine`.


kboyarshinov (approved):
LGTM 👍

PR #340, opened 2023-03-23
use our convention plugins to build the project

gabrielittner:
Significantly reduces the manual Gradle config needed and as a nice side effect we get `tvosSimulatorArm64` and `watchosX64` as multiplatform targets for state machine.


deeplinks-plugin/deeplinks-plugin.gradle.kts
```diff
@@ -1,6 +1,6 @@
 plugins {
     alias(libs.plugins.fgp.gradle)
-//    alias(libs.plugins.publish)
+    // alias(libs.plugins.fgp.publish)
```
fraherm, line 3: Should we remove this line?

fraherm (approved):
LGTM

deeplinks-plugin/deeplinks-plugin.gradle.kts
gabrielittner: No, it's just disabled until it is ready

PR #341, opened 2023-03-23
build with jdk 19

gabrielittner:
Also removed the matrix config since we are just building with one jdk version.


fraherm (approved):
LGTM

PR #369, opened 2023-04-10
implement Stack which represents a single back stack

gabrielittner:
`Stack` exposes some simple functions to mutate the stack like `push`, `pop`, `popUpTo` and `clear`. The backs stack itself is internal and only `computeVisibleEntries` is there to expose the visible part of the back stack which can then be used to show UI on the screen. The stack contains `StackEntry` which have an unique id, the destination that is shown for the entry and the route that was used to navigate there. 

Saving and restoring the state is handled by writing the ids and routes to a `Bundle` and then using that to create new entries from it.

### DIFF
--- a/navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/Stack.kt
+++ b/navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/Stack.kt
@@ -49,7 +49,7 @@ internal class Stack private constructor(
             }
         }
 
-        throw IllegalStateException("Stack did not contain a ScreenDestination $stack")
+        error("Stack did not contain a ScreenDestination $stack")
     }
 
     fun push(route: NavRoute) {
--- a/navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/Stack.kt
+++ b/navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/Stack.kt
@@ -1,0 +1,139 @@
+package com.freeletics.mad.navigator.compose.internal
+
+import android.os.Bundle
+import androidx.core.os.bundleOf
+import com.freeletics.mad.navigator.BaseRoute
+import com.freeletics.mad.navigator.NavRoot
+import com.freeletics.mad.navigator.NavRoute
+import com.freeletics.mad.navigator.compose.ContentDestination
+import com.freeletics.mad.navigator.compose.ScreenDestination
+import com.freeletics.mad.navigator.internal.DestinationId
+import com.freeletics.mad.navigator.internal.destinationId
+import java.util.UUID
+
+internal class Stack private constructor(
+    initialStack: List<StackEntry<*>>,
+    private val destinations: List<ContentDestination<*>>,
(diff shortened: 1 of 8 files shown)


navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/Stack.kt
```diff
@@ -0,0 +1,139 @@
...
+import com.freeletics.mad.navigator.internal.destinationId
+import java.util.UUID
+
+internal class Stack private constructor(
+    initialStack: List<StackEntry<*>>,
+    private val destinations: List<ContentDestination<*>>,
+    private val onStackEntryRemoved: (StackEntry.Id) -> Unit,
+    private val idGenerator: () -> String,
+) {
+    private val stack = ArrayDeque<StackEntry<*>>(20).also {
```
kboyarshinov, line 20: Is there a reason for size `20`?

navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/StackEntry.kt
```diff
@@ -0,0 +1,25 @@
...
+internal class StackEntry<T : BaseRoute>(
+    val id: Id,
+    val route: T,
+    val destination: ContentDestination<T>,
+){
+    val destinationId
+        get() = route.destinationId
+
+    val removable
+        get() = route is NavRoute
```
kboyarshinov, line 21: wouldn't it be more explicit to check if `route !is NavRoot` instead?

kboyarshinov (approved):
LGTM :+1:

navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/Stack.kt
gabrielittner: I just chose a number that should be big enough for most back stacks but also not too big.

navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/Stack.kt
```diff
@@ -0,0 +1,139 @@
...
+                val expectedSize = stack.size - iterator.nextIndex()
+                return ArrayList<StackEntry<*>>(expectedSize).apply {
+                    while (iterator.hasNext()) {
+                        add(iterator.next())
+                    }
+                }
+            }
+        }
+
+        throw IllegalStateException("Stack did not contain a ScreenDestination $stack")
```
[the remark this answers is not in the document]
gabrielittner: [a76b651](https://github.com/freeletics/mad/pull/369/commits/a76b651759c5b203104973e9af5282e1b746ab61)

navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/StackEntry.kt
gabrielittner: Changed it to a when [fc043a6](https://github.com/freeletics/mad/pull/369/commits/fc043a65903ad9fd31fb275aa3ee14603607ea13)

PR #372, opened 2023-04-12
implement MultiStack which handles multiple stacks

gabrielittner:
Building on top of the `Stack` from #369 this supports having multiple back stacks. Generally this exposes the same API as `Stack` and then delegates to one of multiple stacks. It also takes care of creating or removing a stack when needed. The `startStack` is the `Stack` that is created for the start destination of the `NavHost` and our implementation will never remove that stack and also has the root of the start stack as the parent of all other stacks.


navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/MultiStack.kt
```diff
@@ -0,0 +1,181 @@
...
+            onStackEntryRemoved: (StackEntry.Id) -> Unit,
+            idGenerator: () -> String = { UUID.randomUUID().toString() },
+        ): MultiStack {
+            val allStackBundles = bundle.getParcelableArrayList<Bundle>(SAVED_STATE_ALL_STACKS)!!
+            val currentStackId = bundle.getSerializable(SAVED_STATE_CURRENT_STACK)
+            val allStacks = allStackBundles.mapTo(ArrayList(allStackBundles.size)) {
+                Stack.fromState(it, destinations, onStackEntryRemoved, idGenerator)
+            }
+            val startStack = allStacks.find { it.id == root.destinationId }!!
+            val currentStack = allStacks.find { it.id.route.java == currentStackId }!!
```
hoc081098, line 167: How about `first { ... }` is instead of `find  { ... }!!` 😁

gabrielittner: Yes, will update it 👍

gabrielittner: [c050a9c](https://github.com/freeletics/mad/pull/372/commits/c050a9cad75665333d284415eb31a2e86278f623)

navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/MultiStack.kt
```diff
@@ -0,0 +1,181 @@
...
+
+        val newStack = Stack.createWith(root, destinations, onStackEntryRemoved, idGenerator)
+        allStacks.add(newStack)
+        return newStack
+    }
+
+    private fun switchToBackStack(
+        stack: Stack,
+        restoreRootState: Boolean,
+        saveCurrentRootState: Boolean
```
kboyarshinov, line 66: I think `removeCurrentRootState` would be clearer here since it emphasizes positive action instead of `saveCurrentRootState` which is only does something when it's `false`. 

Similar thing about `restoreRootState` since it's not about restoring but rather not clearing. Perhaps `clearNewRootState` would work better.

kboyarshinov, line 66: General question: Do you want to support all 4 combinations of these values? Is there a use case for all of them?

navigator/runtime-experimental/src/test/kotlin/com/freeletics/mad/navigator/compose/internal/MultiStackTest.kt
```diff
@@ -0,0 +1,736 @@
...
+
+        assertThat(removed).isEmpty()
+    }
+
+    @Test
+    fun `visibleEntries is new entry after navigating to root twice`() {
+        val stack = underTest()
+        stack.push(OtherRoot(1), restoreRootState = true, saveCurrentRootState = true)
+        stack.push(SimpleRoot(1), restoreRootState = false, saveCurrentRootState = false)
+        stack.push(OtherRoot(1), restoreRootState = true, saveCurrentRootState = true)
```
kboyarshinov, line 201: I think this API is hard to follow and understand end result. Maybe having separate methods for each use case would make it clearer. This is kind of related to general question if we need to support permutations of boolean parameters. What do you think?

navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/MultiStack.kt
```diff
@@ -0,0 +1,181 @@
...
+                "Can't navigate back from the root of the start back stack"
+            }
+            switchToBackStack(startStack, restoreRootState = false, saveCurrentRootState = false)
+        } else {
+            currentStack.pop()
+        }
+        updateVisibleDestinations()
+    }
+
+    fun <T : BaseRoute> popBackTo(
```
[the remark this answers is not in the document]
gabrielittner: [3bd410a](https://github.com/freeletics/mad/pull/372/commits/3bd410a1d1f4253a5a0a949d3d46799bd2570357)

navigator/runtime-experimental/src/test/kotlin/com/freeletics/mad/navigator/compose/internal/MultiStackTest.kt
kboyarshinov: Thanks for the explanation.

> So we never have restoreRootState = true, saveCurrentRootState: Boolean = false. We could make resetToRoot a separate API which is the case where both are false

I like that, I think it would be good to have it this way.

gabrielittner: Split it up here [30ad593](https://github.com/freeletics/mad/pull/372/commits/30ad5932e5316346f1387f5adf70b5fc8db304b8)

kboyarshinov (approved):
LGTM :+1:

PR #373, opened 2023-04-12
add internal API for NavHost to MultiStackNavigationExecutor

gabrielittner:
These are needed to implement `NavHost`, adding them now so that this can be parallelized with the stack/executor implementation.


kboyarshinov (approved):
LGTM

PR #377, opened 2023-04-16
implement global StoreViewModel

gabrielittner:
This will be used by `MultiStackNavigationExecutor` to store a `SavedStateHandle` and `NavigationExecutorStore` for each `StackEntry` that needs one. This will be a single `Activity` level (or whatever else the parent of the `NavHost` is) `ViewModel` for all screens/entries compared to having one per screen like the AndroidX navigation based implementations have. `removeEntry` will be called manually on backstack changes through the `onStackEntryRemoved` callback which will clear stores and handles once they are unused. 

The `SavedStateHandle.createHandle(...)` and `savedStateHandle.savedStateProvider()` calls are both restricted on the AndroidX side. It would theorectically be possible to have handles without accessing those 2 but that would require having an entry specific `ViewModelStore` (ok to implement and I had this already, but it's overkill for what we need so I removed it again and settled for this solution) as well as a `SavedStateRegistryOwner` and the latter would then also require implementing a `LifecycleOwner` and managing a lifecycle for each entry.

### DIFF
--- a/navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/StoreViewModel.kt
+++ b/navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/StoreViewModel.kt
@@ -16,24 +16,17 @@ internal class StoreViewModel(
     private val savedStateHandles = mutableMapOf<StackEntry.Id, SavedStateHandle>()
 
     fun provideStore(id: StackEntry.Id): NavigationExecutor.Store {
-        var store = stores[id]
-        if (store == null) {
-            store = NavigationExecutorStore()
-            stores[id] = store
-        }
-        return store
+        return stores.getOrPut(id) { NavigationExecutorStore() }
     }
 
     @SuppressLint("RestrictedApi")
     fun provideSavedStateHandle(id: StackEntry.Id): SavedStateHandle {
-        var savedStateHandle = savedStateHandles[id]
-        if (savedStateHandle == null) {
+        return savedStateHandles.getOrPut(id) {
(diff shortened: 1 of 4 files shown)


navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/internal/StoreViewModel.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val stores = mutableMapOf<StackEntry.Id, NavigationExecutorStore>()
+    private val savedStateHandles = mutableMapOf<StackEntry.Id, SavedStateHandle>()
+
+    fun provideStore(id: StackEntry.Id): NavigationExecutor.Store {
+        var store = stores[id]
+        if (store == null) {
+            store = NavigationExecutorStore()
+            stores[id] = store
+        }
+        return store
```
hoc081098, line 24: Imo, `stores.getOrPut { }` is more concise

gabrielittner: thanks, I often forget about many of the collection functions in the stdlib 👍 [9225227](https://github.com/freeletics/mad/pull/377/commits/922522757030a159dc610a6f6dbe6417b6f7544b)

hoc081098: 😁 There are about 200 of them

kboyarshinov (approved):
LGTM 👍

PR #378, opened 2023-04-16
implement ActivityStarter

gabrielittner:
Just a simple abstraction that handles navigating with an `ActivityRoute`. The main reason for having this as it's own class is that it requires a `Context` so I want to isolate it.


kboyarshinov (approved):
👍

PR #381, opened 2023-04-18
NavHost: system back handling

gabrielittner:
Adds a callback to the `OnBackPressedDispatcher` and enables/disables it based on the state exposed by the executor.

### DIFF
--- a/navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/NavHost.kt
+++ b/navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/NavHost.kt
@@ -70,7 +70,10 @@ public fun NavHost(
 
 @Composable
 private fun SystemBackHandling(executor: MultiStackNavigationExecutor) {
-    val backPressedDispatcher = LocalOnBackPressedDispatcherOwner.current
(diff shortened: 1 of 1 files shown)


navigator/runtime-experimental/src/main/java/com/freeletics/mad/navigator/compose/NavHost.kt
```diff
@@ -59,6 +68,31 @@ public fun NavHost(
...
+
+        }
+    }
+    LaunchedEffect(executor, callback) {
+        snapshotFlow { executor.canNavigateBack.value }
+            .distinctUntilChanged()
+            .collect { callback.isEnabled = it }
+    }
+    DisposableEffect(backPressedDispatcher, callback) {
+        backPressedDispatcher!!.onBackPressedDispatcher.addCallback(callback)
```
kboyarshinov, line 88: I think this should fail early in case `LocalOnBackPressedDispatcherOwner` can't be obtained

gabrielittner: [5fccfae](https://github.com/freeletics/mad/pull/381/commits/5fccfaee40b6c895c20c0fecf03ea14220c56695)

kboyarshinov (approved):
:shipit:

PR #383, opened 2023-04-18
split resetToRoot from navigateToRoot

gabrielittner:
As discussed in #372.

One thing that I don't like entirely is that for the AndroidX navigation implementation there isn't anything that enforces that `resetToRoot` is called with the current root or the start destination root (that one is always the parent). Not a big issue though and it would still result in a valid back stack, it's just that in that case navigateToRoot(..., restore = false) would be the better choice. In `MultiStack` we can enforce this ourselves.


kboyarshinov (approved):
LGTM :+1:

PR #384, opened 2023-04-18
NavHost: destination changed callback

gabrielittner:
Uses the top visible entry and invokes the callback when it changes.


kboyarshinov (approved):
LGTM :+1:

PR #385, opened 2023-04-18
NavHost: support comopose saveable on a per destination basis

gabrielittner:
Adds support for `rememberSaveable` which will get cleared when a screen is removed from the back stack.


kboyarshinov (approved):
:+1:

PR #386, opened 2023-04-18
NavHost: support rendering dialog and bottom sheet destinations

gabrielittner:
Properly shows dialogs and bottom sheets. Similar to AndroidX navigation we currently only support showing one bottom sheet at a time.


kboyarshinov (approved):
LGTM :+1:

PR #397, opened 2023-04-26
implement MultiStackNavigationExecutor

gabrielittner:
The last part to get `navigator-experimental` generally working. Most of the implementation is just delegating to already added components like `MultiStack` and the store. The main new addition is the deep link handling.


kboyarshinov (approved):
:shipit:

PR #417, opened 2023-05-26
remember the state machine and other injected dependencies

gabrielittner:
Avoids creating new instances of each injected object on each recomposition when they are not scoped.


kboyarshinov (approved):
:shipit:

PR #433, opened 2023-06-26
various sample fixes/improvements

gabrielittner:
- add some gradle.properties to avoid running out of memory and get rid of warnings
- fix build health in the sample app
- change bottom sheet to not use `fillMaxSize`
- use an included build to directly have the local library code in the sample


fraherm (approved):
LGTM

PR #434, opened 2023-06-26
more sample screens

gabrielittner:
- rename :feature:a to feature:root
- add :feature:screen
- feature screen can navigate to itself as well as to the bottom sheet


fraherm (approved):
Nice

PR #436, opened 2023-06-27
compose: replace dialog and bottom sheet destinations with overlay

gabrielittner:
We now have an overlay destination type. The difference compared to dialog and bottom sheet is that this destination type does not add a container like `Dialog` or `ModalBottomSheetLayout` anymore, it will just show the content on top of screen destinations. This moves the responsibility of adding the container to the destinations. For the simple case it's a bit more effort but it gives a lot more flexibility as well since you can customize the bottom sheet or dialog on a per destination basis. It also means we don't need all the bottom sheet parameters on `NavHost` anymore and we can drop the dependency on accompanist-navigation-material.

This fixes the navigation issues that we had with bottom sheets (dismissing it by clicking outside or dragging it closed does not pop the back stack) and dialogs (not displayed).

For the AndroidX navigation implementation this ads `OverlayNavigator` and `OverlayNavHost` (internally). These are a mostly 1:1 copy of `DialogNavigator` and `DialogHost` from AndroidX navigation. The only change to them is that `DialogHost` does not add a wrapper `Dialog` around `destination.content(backStackEntry)`.

In the experimental implementation the change is to just show all visible destinations in the same way instead of wrapping some based on the type.

### DIFF
--- a/docs/navigator/get-started.md
+++ b/docs/navigator/get-started.md
@@ -86,17 +86,20 @@ like this:
     that was passed as the generic type parameter. The lambda function then gets an instance of that
     `NavRoute` and calls the `@Composable` function that should be shown.
 
-    The other 2 functions to create destinations are `DialogDestination` and `BottomSheetDestination` -
-    they declare destinations that use a dialog or bottom sheet as a container instead of being shown
-    full screen.
+    There is also an `OverlayDestination` function to declare destinations that use a dialog or bottom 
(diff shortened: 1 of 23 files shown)


fraherm (approved):

fraherm:
Great work! 🚀

kboyarshinov (approved):
🚢

PR #442, opened 2023-06-30
MAD -> Khonshu

gabrielittner:
New name 🎉 

Since the `StateMachine` interface is used by flow redux (and more) I've added a `state-machine-legacy` artifact that has the interface with the old package name. That interface extends the one with the new package name. This allows us to do one release where still publish the "mad StateMachine" and then flow redux can be changed to the new interface without it being a breaking change. After the next release `state-machine-legacy` can be deleted.


kboyarshinov (approved):
👍 🐦

PR #459, opened 2023-07-17
add @Language annotation to test code

gabrielittner:
Makes reading the code in strings nicer since this adds highlighting to it. The suppresses are for not valid warnings that this highlighting shows in those strings.

Also improved the naming and structure of the test helpers a bit.


fraherm (approved):
Nice 👍🏻

PR #463, opened 2023-07-18
make it possible to directly use generated component as parent

gabrielittner:
For now this uses the existing mechanism that we have for `NavEntryComponent` through `NavEntryComponentGetter` and `NavDestinationComponent`. The next 2 follow ups are first deleting `NavEntryComponent` and then improving the look up mechanism to be much simpler. However to keep changes here smaller it makes more sense to use the existing mechanism for now.

`NavDestinationComponent` has a `Map<Class<*>, ComponentProvider<*, *>`. We provide the generated `ComponentProvider` into this using the `scope` (in practice a route class) to it. We're also contributing `NavDestinationComponent` to the destination scope that we can use that for the look up. This now allows screen with `ExampleARoute` to use `ExampleParentRoute` as parent scope. The component will be looked up by getting `NavDestinationComponent` and retrieving the component provider from the map by using `ExampleParentRoute` as key.

### DIFF
--- a/codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/codegen/nav/NavDestinationModuleGenerator.kt
+++ b/codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/codegen/nav/NavDestinationModuleGenerator.kt
@@ -3,14 +3,22 @@ package com.freeletics.khonshu.codegen.codegen.nav
 import com.freeletics.khonshu.codegen.BaseData
 import com.freeletics.khonshu.codegen.Navigation
 import com.freeletics.khonshu.codegen.codegen.Generator
+import com.freeletics.khonshu.codegen.codegen.common.componentProviderClassName
 import com.freeletics.khonshu.codegen.codegen.common.composableName
 import com.freeletics.khonshu.codegen.codegen.fragment.fragmentName
+import com.freeletics.khonshu.codegen.codegen.util.componentProvider
 import com.freeletics.khonshu.codegen.codegen.util.contributesToAnnotation
+import com.freeletics.khonshu.codegen.codegen.util.intoMap
 import com.freeletics.khonshu.codegen.codegen.util.intoSet
 import com.freeletics.khonshu.codegen.codegen.util.module
+import com.freeletics.khonshu.codegen.codegen.util.navComponentProvider
+import com.freeletics.khonshu.codegen.codegen.util.optInAnnotation
 import com.freeletics.khonshu.codegen.codegen.util.provides
+import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.CodeBlock
 import com.squareup.kotlinpoet.FunSpec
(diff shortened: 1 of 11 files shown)


codegen/src/main/kotlin/com/freeletics/khonshu/codegen/internal/ComponentProvider.kt
```diff
@@ -48,11 +47,37 @@ public inline fun <reified C : Any, P : Any, R : BaseRoute> component(
     executor: NavigationExecutor,
     context: Context,
     parentScope: KClass<*>,
+    // TODO remove parameter
+    @Suppress("UNUSED_PARAMETER") destinationScope: KClass<*>,
```
gabrielittner, line 51: Will be removed in the mentioned follow up

codegen/src/main/kotlin/com/freeletics/khonshu/codegen/internal/ComponentProvider.kt
```diff
@@ -48,11 +47,37 @@ public inline fun <reified C : Any, P : Any, R : BaseRoute> component(
...
+
+/**
+ * Creates a [ViewModel] for the given [destinationId]. The `ViewModel.Factory` will use [parentScope]
+ * to lookup a parent component instance. That component will then be passed to the given [factory]
+ * together with a [SavedStateHandle] and the passed in [destinationId].
+ *
+ * To be used in generated code.
+ */
+@InternalCodegenApi
+@JvmName("componentForNavDestination") // TODO give unique name to method
```
gabrielittner, line 69: Same here.

codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/codegen/nav/NavDestinationModuleGenerator.kt
```diff
@@ -60,4 +70,25 @@ internal class NavDestinationModuleGenerator(
...
+            .addAnnotation(mapKeyAnnotation())
+            .returns(
+                componentProvider.parameterizedBy(
+                    STAR,
+                    STAR,
+                ).copy(annotations = listOf(AnnotationSpec.builder(JvmSuppressWildcards::class).build())),
+            )
+            .addStatement("return %T", componentProviderClassName)
+            .build()
+    }
```
fraherm, line 87: @gabrielittner Do you think it would make sense to add generated code example in comments above the function? IMO that would make it easier to read.

fraherm (approved):

codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/codegen/nav/NavDestinationModuleGenerator.kt
gabrielittner: I can make sense but it wouldn't be maintainable. Even the tests are annoying to update for some tests

PR #475, opened 2023-07-28
added multinav event

josirichter:
This adds a MultiNavEvent to combine multiple navigation calls from one navigator. 

When triggering multiple navigation events from a navigator it would sometimes happen that only the first one would actually be triggered. This would combine the nav events into ensuring that all of them are followed through.

### DIFF
--- a/navigation/src/main/kotlin/com/freeletics/khonshu/navigation/NavEventNavigator.kt
+++ b/navigation/src/main/kotlin/com/freeletics/khonshu/navigation/NavEventNavigator.kt
@@ -10,9 +10,12 @@ import com.freeletics.khonshu.navigation.internal.NavEvent
 import com.freeletics.khonshu.navigation.internal.NavEvent.ActivityResultEvent
 import com.freeletics.khonshu.navigation.internal.NavEvent.BackEvent
 import com.freeletics.khonshu.navigation.internal.NavEvent.BackToEvent
+import com.freeletics.khonshu.navigation.internal.NavEvent.MultiNavEvent
 import com.freeletics.khonshu.navigation.internal.NavEvent.NavigateToActivityEvent
 import com.freeletics.khonshu.navigation.internal.NavEvent.NavigateToEvent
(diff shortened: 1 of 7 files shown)


navigation/src/main/kotlin/com/freeletics/khonshu/navigation/NavEventNavigator.kt
```diff
@@ -161,6 +164,17 @@ public open class NavEventNavigator {
         sendNavEvent(event)
     }
 
+    /**
+     * Triggers a new [NavEvent] that collects and combines the nav events sent in the block so they can be
+     * handled individually.
+     *
+     * Note: This should be used when navigating multiple times fe calling `navigateBackTo` followed by `navigateTo`.
```
gabrielittner, line 171:
```suggestion
     * Note: This should be used when navigating multiple times, for example calling `navigateBackTo` followed by `navigateTo`.
```

navigation/src/main/kotlin/com/freeletics/khonshu/navigation/internal/NavEventCollector.kt
```diff
@@ -0,0 +1,51 @@
...
+public interface Navigator {
+    public fun navigateTo(route: NavRoute)
+    public fun navigateToRoot(root: NavRoot, restoreRootState: Boolean = false)
+    public fun navigateUp()
+    public fun navigateBack()
+    public fun <T : NavRoute> navigateBackTo(navRoute: KClass<T> , inclusive: Boolean = false)
+    public fun resetToRoot(root: NavRoot)
+}
+
+internal class NavEventCollector : Navigator {
```
gabrielittner, line 17: I think we can just merge these 2 into one class. The class can be public as long as we make the constructor and the `navEvents` property `internal`. That would also using a reified inline function for `navigateBackTo` like we do in `NavEventNavigator`

gabrielittner (approved):
Looks good 👍  just left 2 small comments.

For the build to succeed you need to run `./gradlew apiDump` once since there were changes to the public API of the library.

PR #477, opened 2023-07-28
remove legacy state machine module

gabrielittner:
It's been released once which is enough and FlowRedux the main consumer also now has a release using that version.


kboyarshinov (approved):
👍

PR #482, opened 2023-08-04
Multinav test adaptions

josirichter:
I added a test method for the NavigatorTurbine to combine nav events in tests and properly test MultiNavEvents.


gabrielittner (approved):
:shipit:

PR #492, opened 2023-08-10
update naming of ComposableParameter functions

gabrielittner:
- instead of having `getSendActionParameter` and `getStateParameter` with basically the same implementation `getParameterWithType`
- rename `getComposeParameters` to `getInjectedParameters` to make its behavior clearer and also share the logic for it with `getParameterWithType`


kboyarshinov (approved):
👍

PR #493, opened 2023-08-10
rewrite type parameter resolution to do a single pass

gabrielittner:
We have this code to resolve the type parameters of `StateMachine` from a concrete implementation, which we then use to identify the state and sendAction parameters in a compose function. Right now we're first collecting type references for all super types and then we go through that twice to actually resolve the type parameters. While looking into a solution for this in KSP I came up with a better implementation. With this it only requires a single pass instead of 3. More description is directly on the code.


kboyarshinov (approved):
Nice :+1:

PR #494, opened 2023-08-10
unify test helpers

gabrielittner:
Creates a `KhonshuCompilation` abstraction that currently has 2 modes
- just compile the given code without any compiler plugins
- compile with anvil running

This basically matches the 2 tests we do in `TestHelper`. Now this is used also for the `AllSuperTypesTest`. This can be easily extended to add a ksp mode afterwards.


kboyarshinov (approved):
LGTM 👍

PR #495, opened 2023-08-11
change navigation to be a multiplatform module

gabrielittner:
The goal here is not to actually support multiplatform (at least not in the short term). KSP's main entry point for us is `getSymbolsWithAnnotation` which will get the annotation class as parameter and to read the annotation from a symbol we again need the class and that will return an instance of the annotation. This is nice since we can then just access properties of the annotations and default values just work (unlike Anvil where we need the names of properties as strings) but it also means that the annotations need to be available in a jvm module. The annotations reference `BaseRoute` which means that that also needs to be available in a jvm module. So the goal of this PR is to make `BaseRoute` available for jvm only code.

Using the multiplatform plugin in this case has a few advantages:
- it stays a single module
- expect/actual allows us to make parcelable/parcelize work
- we could also move some of the deep link code to common that with the gradle plugin

The next pr will be to also change `codegen` with the annotations to a multiplatform module.


kboyarshinov (approved):
LGTM 👍

PR #500, opened 2023-08-16
change codegen to be a multiplatform module

gabrielittner:
Same as #495 but for the codegen module.

Instead of doing this for 3 modules I'm just merging the codegen-fragment and codegen-compose modules into this one. The only 2 dependecies that were specific to the project type were Renderer and Fragments, these are added as compileOnly. For these 2 we also need an expect/actual. It's not great but it's temporary until we get rid of fragments which we're currently planning, so I consider it acceptable for now.

The scope annotations have also been moved to `codegen` since the only reason for an extra module was to have them available in non Android modules.

I'm keeping the 3 old and now empty modules for one release to make updating easier.


kboyarshinov (approved):
:shipit:

PR #502, opened 2023-08-17
use DestinationType in compiler

gabrielittner:
Now that the compiler can depend on the runtime we can directly use the `DestinationType` enum.


kboyarshinov (approved):
Nice 👍

PR #504, opened 2023-08-18
add KSP processor and implement it for compose

gabrielittner:
First half of the KSP implementation, the fragment based compose and renderer annotations will be handled in the next PR.

PR #506, opened 2023-08-21
support KSP in Fragment codegen

gabrielittner:
Adds the parts of codegen to KSP that #504 omitted. With this everything can be generated through KSP.


kboyarshinov (approved):
:shipit:

PR #512, opened 2023-08-24
improve naming

gabrielittner:
Primarily renames `ComposeScreenGenerator` to `OuterComposableGenerator` and `ComposeGenerator` to `InnerComposableGenerator` to hopefully make it a bit clearer what they do and how they relate to each other. Also added comments for that.

Also removed some unused code.


codegen-compiler-test/src/test/kotlin/com/freeletics/khonshu/codegen/codegen/TestHelper.kt
```diff
@@ -23,11 +23,9 @@ internal fun test(data: BaseData, fileName: String, source: String, expectedCode
...
 
 private fun compile(fileName: String, source: String, data: BaseData, expectedCode: String) {
-    val generatedCode = when (data) {
-        is ComposeFragmentData -> FileGenerator().generate(data).toString()
-        is ComposeScreenData -> FileGenerator().generate(data).toString()
-        is RendererFragmentData -> FileGenerator().generate(data).toString()
-    }
+    val generatedCode = FileGenerator().generate(data).toString()
+
+    assertThat(generatedCode).isEqualTo(expectedCode)
```
gabrielittner, line 28: Moved this up since it's helpful to first check whther the output matches what we expect and only after that check if it compiles.

josirichter (approved):

PR #513, opened 2023-08-25
parse annotation parameters indirectly

gabrielittner:
It turns out that we need to manually parse parameters from `KSAnnotation` instead of using `getAnnotationsByType` to get an instance of the annotation after all. The latter only works for primitive values and for classes that have already been compiled (so classes coming from dependencies). This is why it works in tests because we use an already compile state machine class but it wouldn't work in real projects where the state machine is defined in the same module as the annotated composable/renderer.

Can be tested by setting the Gradle property `fgp.kotlin.khonshuKsp=true` to enable KSP for the sample project, e.g. `./gradlew -p sample/simple/ assemble -Pfgp.kotlin.khonshuKsp=true`


kboyarshinov (approved):
:shipit:

PR #520, opened 2023-09-06
generate Activity for NavHostActivityData

gabrielittner:
First new part of the NavHostActivity codegen. This generates the `Activity` subclass which will obtain the component and then call the already generated wrapper composable from `setContent`. `NavHostActivityData` gets a property for the activity base class (there already is a parameter for this on the annotation).

### DIFF
--- a/codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/codegen/compose/ActivityGenerator.kt
+++ b/codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/codegen/compose/ActivityGenerator.kt
@@ -1,0 +1,64 @@
+package com.freeletics.khonshu.codegen.codegen.compose
+
+import com.freeletics.khonshu.codegen.BaseData
+import com.freeletics.khonshu.codegen.NavHostActivityData
+import com.freeletics.khonshu.codegen.codegen.Generator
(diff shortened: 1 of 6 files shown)


codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/codegen/compose/ActivityGenerator.kt
```diff
@@ -0,0 +1,64 @@
...
+
+internal val Generator<out BaseData>.fragmentName
+    get() = "Khonshu${data.baseName}Activity"
+
+internal class ActivityGenerator(
+    override val data: NavHostActivityData,
+) : Generator<NavHostActivityData>() {
+
+    internal fun generate(): TypeSpec {
+        return TypeSpec.classBuilder(fragmentName)
```
kboyarshinov, line 29:
```suggestion
internal val Generator<out BaseData>.activityName
    get() = "Khonshu${data.baseName}Activity"

internal class ActivityGenerator(
    override val data: NavHostActivityData,
) : Generator<NavHostActivityData>() {

    internal fun generate(): TypeSpec {
        return TypeSpec.classBuilder(activityName)
```

kboyarshinov (approved):
👍

PR #528, opened 2023-09-07
NavHost codegen

gabrielittner:
We expect all annotated composables to have a `navHost: @Composable (NavRoot, ((BaseRoute) -> Unit)?) -> Unit` parameter (parsing is not implemented yet, so it's only in the model for now). We then add that parameter to the generated composable and let our generated `Activity` pass in the composable for it. That composable will call our `NavHost` and automatically obtains destinations and deep links from the component. The 2 composable parameters are the start root and and optional destination change callback.


kboyarshinov (approved):
LGTM 👍

kboyarshinov (approved):
LGTM 👍

PR #530, opened 2023-09-08
ActivityModule codegen

gabrielittner:
Generates a module that has multi bindings for `DeepLinkHandler` and `DeepLinkHandler.Prefix` for apps that don't have deep links. I didn't add one for `NavDestination` because the app would not work if there are no contributed destinations.


kboyarshinov (approved):
👍

PR #531, opened 2023-09-08
NavHostActivity: implement Anvil parsing

gabrielittner:
Parses the annotation into `NavHostActivityData` for Anvil.  The output stays unchanged and the tests also run with Anvil compilation now. KSP will be in a follow up PR.


kboyarshinov (approved):
👍

PR #536, opened 2023-09-12
update sample

gabrielittner:
Uses the new `@NavHostActivity` in the same to generate the `Activity` and take care of the general set up.

Also updates the sample to use Anvil's new scope annotations instead of ours (missed to do that when updating codegen itself).


kboyarshinov (approved):
👍

PR #537, opened 2023-09-13
remove non navigation codegen

gabrielittner:
Removes the old annotations that are not navigation related. The Fragment/Compose codegen now always generates code that includes the destinations and navigation set up.


kboyarshinov (approved):
:shipit:

PR #542, opened 2023-09-22
adapt file generation for compose fragments

josirichter:
I've changed the file generation for compose fragments to include "Fragment" in the name. This. will allow for generating both the compose only and fragment setup if needed.

### DIFF
--- a/codegen-compiler-test/src/test/kotlin/com/freeletics/khonshu/codegen/codegen/FileGeneratorTestComposeFragment.kt
+++ b/codegen-compiler-test/src/test/kotlin/com/freeletics/khonshu/codegen/codegen/FileGeneratorTestComposeFragment.kt
@@ -28,7 +28,7 @@ internal class FileGeneratorTestComposeFragment {
     )
 
     private val data = ComposeFragmentData(
-        baseName = "Test",
+        baseName = "FragmentTest",
         packageName = "com.test",
         scope = ClassName("com.test", "TestScreen"),
         parentScope = ClassName("com.test.parent", "TestParentScope"),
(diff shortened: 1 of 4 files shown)


codegen-compiler-test/src/test/kotlin/com/freeletics/khonshu/codegen/codegen/FileGeneratorTestComposeFragment.kt
```diff
@@ -185,7 +185,7 @@ internal class FileGeneratorTestComposeFragment {
 
         """.trimIndent()
 
-        test(data, "com/test/Test.kt", source, expected)
+        test(data, "com/test/FragmentTest.kt", source, expected)
```
gabrielittner, line 188: This one should stay as `Test.kt` as it's the file name for the sources, not for the generated code. Same for the other tests below.

gabrielittner (approved):
One small comment, otherwise looks good 👍

codegen-compiler-test/src/test/kotlin/com/freeletics/khonshu/codegen/codegen/FileGeneratorTestComposeFragment.kt
gabrielittner: Ah, we're computing the generated file name based on this as well https://github.com/freeletics/khonshu/blob/main/codegen-compiler/src/testFixtures/kotlin/com/freeletics/khonshu/codegen/KhonshuCompilation.kt#L154-L158. So let's keep it as is

PR #544, opened 2023-09-23
SystemBackHandling: remove unnecessary distinctUntilChanged

hoc081098:
- Ref: https://developer.android.com/reference/kotlin/androidx/compose/runtime/package-summary#snapshotFlow(kotlin.Function0)
> If the result of [block](https://developer.android.com/reference/kotlin/androidx/compose/runtime/package-summary#snapshotFlow(kotlin.Function0)) is not [equal to](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/equals.html) the previous result, the flow will emit that new result. (This behavior is similar to that of [Flow.distinctUntilChanged](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/index.html).)


gabrielittner (approved):
Thank you

PR #545, opened 2023-09-25
rename @ComposeDestination annotation

gabrielittner:
If we want to use them in parallel having the same name is inconvenient and also makes it hard to talk about them. This renames the compose only annotation to `@NavDestination` (matches `@NavHostActivity` and once fragment support is gone these will be the only annotations codegen has) and the fragment one is `@ComposeFragmentDestination`.


josirichter (approved):
good idea 👍

PR #566, opened 2023-10-06
move as much DeepLink code as possible to common

gabrielittner:
The goal is to let the Gradle plugin access some of this so that we're not duplicating logic. For that we need some parts in the common source set. I've turned the constructors for `DeepLink` into top level functions with the same name so that the ones that use `ActivityRoute` can stay in the Android source set. The methods to turn a `DeepLink` into an `Intent` and other Android objects are now extension functions.


kboyarshinov (approved):
LGTM 👍

PR #574, opened 2023-10-17
add deep links test helper

gabrielittner:
Adds a `DeepLinkDefinitions.containsAllDeepLinks` method to check that all deep links loaded from the TOML file have a `DeepLinkHandler` and the other way around. 

Closes #572


kboyarshinov (approved):
👍

PR #580, opened 2023-10-20
fix: error "Navigator ... is replacing an already attached ..."

hoc081098:
When changing `startRoute`, that causes `navController.navigatorProvider.addNavigator(CustomActivityNavigator(context))` is called again.
`check(previousNavigator?.isAttached != true)` don't passes.

```kotlin
package androidx.navigation

public open class NavigatorProvider {
 @CallSuper
    public open fun addNavigator(
        name: String,
        navigator: Navigator<out NavDestination>
    ): Navigator<out NavDestination>? {
        ...
        val previousNavigator = _navigators[name]
        if (previousNavigator == navigator) {
            return navigator
        }
        check(previousNavigator?.isAttached != true) {
            "Navigator $navigator is replacing an already attached $previousNavigator"
        }
       ...
    }
}
```

P/S:
Although changing startRoute does not follow [Fixed start destination](https://developer.android.com/guide/navigation/principles?hl=en#fixed_start_destination), but in some cases, we must to do this.

### DIFF
--- a/navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
+++ b/navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
@@ -1,5 +1,7 @@
 package com.freeletics.khonshu.navigation.compose
 
+import androidx.navigation.NavDestination as AndroidXNavDestination
+import androidx.navigation.compose.NavHost as AndroidXNavHost
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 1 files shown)


hoangchungk53qx1:
@hoc081098 💯

gabrielittner:
Thanks for this PR and the other one, I'll take a loot at them this week

gabrielittner (approved):
LGTM. We should probably turn the `Set` parameters into `ImmutableSet` at some point to make he effect and remember more efficient

PR #581, opened 2023-10-22
collectAndHandleNavEvents: collect events in `Dispatcher.Main.immediate`

hoc081098:


navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavEventNavigator.kt
```diff
@@ -37,11 +37,7 @@ public open class NavEventNavigator {
     private val _navEvents = Channel<NavEvent>(Channel.UNLIMITED)
 
     @InternalNavigationApi
-    public val navEvents: Flow<NavEvent> = flow {
-        for (result in _navEvents) {
-            emit(result)
-        }
-    }
+    public val navEvents: Flow<NavEvent> = _navEvents.receiveAsFlow()
```
hoc081098, line 40: `receiveAsFlow` docs:
> Represents the given receive channel as a hot flow and [receives](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-receive-channel/receive.html) from the channel in fan-out fashion every time this flow is collected. One element will be emitted to one collector only.

Although docs used the phrase `hot flow`, the internal implementation is the same as `flow { for(...) emit }` as before

hoc081098: https://github.com/Kotlin/kotlinx.coroutines/blob/ed0cf7aa02b1266cb81e65e61b3a97b0e041a817/kotlinx-coroutines-core/common/src/flow/Channels.kt#L32-L45

gabrielittner (approved):
Thank you 👍

PR #583, opened 2023-10-27
Change the result with receiveAsFlow for Channel

hoangchungk53qx1:
Channel implementation can be changed with receiveAsFlow()
because Chanel is always hotflow.

### DIFF
--- a/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/ResultOwner.kt
+++ b/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/ResultOwner.kt
@@ -83,7 +83,7 @@ public class PermissionsResultRequest internal constructor() :
         /**
          * The app has access to the requested permission.
          */
-        public object Granted : PermissionResult
+        public data object Granted : PermissionResult
 
         /**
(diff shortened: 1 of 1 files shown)


navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/ResultOwner.kt
```diff
@@ -20,17 +21,13 @@ import kotlinx.parcelize.Parcelize
  * to one collector at a time.
  */
 public sealed class ResultOwner<R> {
-    private val _results = Channel<R>(capacity = Channel.UNLIMITED)
 
     /**
-     * Emits any result passed to [onResult]. Results will only be delivered
-     * to one collector at a time.
+     * A [Flow] of results that can be collected by a collector.
```
gabrielittner, line 26: I'd prefer if we keep the previous comment on the `public val results`

gabrielittner (approved):
Thanks 👍

gabrielittner:
For the data object change, it's generally fine, you'd just need to run `./gradlew apiDump` once to update the api file.

PR #584, opened 2023-10-28
apply `data object`

hoc081098:


docs/navigation/activities.md
hoc081098, line 1: It looks like The auto-format of Android-Studio removes trailing-space on each line.

gabrielittner (approved):
Thanks for going through all usages and even the docs

PR #586, opened 2023-10-30
Create `build-simple-sample.yml` workflow

hoc081098:
Build `:app:simple` when pushing to main or opening PR to main


hoangchungk53qx1:
nice ✌️ ,

.github/workflows/build-simple-sample.yml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    name: Build simple sample
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./sample/simple
```
gabrielittner, line 15: We could alternatively use ./gradlew -p sample/simple ...` below and avoid the `../..`

.github/workflows/build-simple-sample.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Install Android SDK
+        run: ../../.github/android-sdk.sh
+
+      - uses: gradle/gradle-build-action@v2
+
+      - name: Build with Gradle
+        run: ./gradlew :app:simple:assembleDebug --stacktrace
+
+      - name: Check code style
+        run: ../../.kotlinw ../../.kts/ktlint.main.kts --fail-on-changes
```
gabrielittner, line 39: The script finds all kt files regardless of Gradle projects, so the existing pr check should be enough

gabrielittner (approved):
Thanks, this is great. Left 2 minor comments but still going ahead with merging this so that it can run

PR #592, opened 2023-11-06
add navController as argument to compose NavHost function

josirichter:
I moved the NavController to the inputs of the NavHost composable so that it can be passed in when using the NavHost with a bottom navigation.


gabrielittner (approved):

navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
```diff
@@ -53,13 +57,17 @@ public fun NavHost(
...
 ) {
     val context = LocalContext.current
 
     val overlayNavigator = remember { OverlayNavigator() }
     val customActivityNavigator = remember(context) { CustomActivityNavigator(context) }
-    val navController = rememberNavController(overlayNavigator, customActivityNavigator)
+
+    navController.navigatorProvider.addNavigator(overlayNavigator)
+    navController.navigatorProvider.addNavigator(customActivityNavigator)
+
```
hoc081098, line 70: Two statements can throw `"Navigator ... is replacing an already attached ..."`. https://github.com/freeletics/khonshu/pull/580 has fixed the similar problem.

hoc081098: I also notice they are `side-effect`s, and should be in Compose Effect(s)

gabrielittner: @hoc081098 I don't think the exception can happen here anymore. `addNavigator` has an early return if the given navigator is the same as the `previousNavigator`. Since both navigators are being remembered they will be the same even after a re-composition. Before your PR the `CustomActivityNavigator` would always be newly instantiated

hoc081098: @gabrielittner I got it, thank you for clarifying 🙏

PR #594, opened 2023-11-06
replace DestinationType with Overlay marker interface

gabrielittner:
Instead of having the `DestinationType` enum to decide whether to generate a `ScreenDestination` or an `OverlayDestination` (for fragments `DialogDestination`) this introduces an `Overlay` marker interface which can be added to any `NavRoute`. If it was added the codegen behaves like `DestinationType.OVERLAY` before.

The advantage of this is that other parts for example a destination change listener can check whether a new destination is an overlay by simply doing `route is Overlay`. We will use that instead of the Fragment class in our listener to decide whether bottom navigation should be shown or not.

### DIFF
--- a/codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/parser/ksp/CommonParser.kt
+++ b/codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/parser/ksp/CommonParser.kt
@@ -5,6 +5,7 @@ import com.freeletics.khonshu.codegen.codegen.util.asLambdaParameter
 import com.freeletics.khonshu.codegen.codegen.util.baseRoute
 import com.freeletics.khonshu.codegen.codegen.util.functionToLambda
 import com.freeletics.khonshu.codegen.codegen.util.navHostLambda
+import com.freeletics.khonshu.codegen.codegen.util.overlay
 import com.freeletics.khonshu.codegen.codegen.util.stateMachine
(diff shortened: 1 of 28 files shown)


codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/parser/ksp/CommonParser.kt
```diff
@@ -104,7 +105,11 @@ internal fun ClassName.stateMachineParameters(resolver: Resolver, logger: KSPLog
 internal fun KSClassDeclaration.findRendererFactory(logger: KSPLogger): ClassName? {
     val factory = declarations.filterIsInstance<KSClassDeclaration>()
         .firstNotNullOfOrNull {
-            if (it.allSuperTypes(false).any { superType -> superType == viewRendererFactory }) {
+            if (it.allSuperTypes(false).any { superType ->
+                    println("test: $superType")
```
josirichter, line 109: nit: do we still need this?

josirichter (approved):
🎉 👍

codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/parser/ksp/CommonParser.kt
gabrielittner: Thanks for catching this, removed it [7d686d5](https://github.com/freeletics/khonshu/pull/594/commits/7d686d5dd8fe52202e30abc6a982bdd255282d26)

PR #603, opened 2023-11-16
Add NavHost overload that accepts NavRoute

kboyarshinov:
Same as existing `NavHost` but accepts `NavRoute`

Depends on #602

### DIFF
--- a/navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
+++ b/navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
@@ -62,7 +62,7 @@ public fun NavHost(
     navController: NavHostController = rememberNavController(),
     destinationChangedCallback: ((BaseRoute) -> Unit)? = null,
 ) {
-    NavHost(
+    InternalNavHost(
         startRoute = startRoute,
         destinations = destinations,
         modifier = modifier,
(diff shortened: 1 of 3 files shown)


navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
```diff
@@ -61,6 +61,44 @@ public fun NavHost(
     deepLinkPrefixes: Set<DeepLinkHandler.Prefix> = emptySet(),
     navController: NavHostController = rememberNavController(),
     destinationChangedCallback: ((BaseRoute) -> Unit)? = null,
+) {
+    NavHost(
+        startRoute = startRoute,
```
gabrielittner, line 66: This is currently a recursive call and not calling the other composable since the shared base type is `BaseRoute`.

gabrielittner: Fixed in [b4c242c](https://github.com/freeletics/khonshu/pull/603/commits/b4c242ceabebf09688676db92ce92d3a04304bc1)

gabrielittner (approved):

navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
```diff
@@ -61,6 +61,66 @@ public fun NavHost(
     deepLinkPrefixes: Set<DeepLinkHandler.Prefix> = emptySet(),
     navController: NavHostController = rememberNavController(),
     destinationChangedCallback: ((BaseRoute) -> Unit)? = null,
+) {
+    InternalNavHost(
```
hoc081098, line 65: We can annotate two functions with `@NonRestartableComposable`, just an optimization.

gabrielittner: 👍  will open a pr for this

navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
kboyarshinov: :man_facepalming:, thanks for the fix :+1:

PR #610, opened 2023-11-20
Add optional executor param to NavigationSetup function

kboyarshinov:
Add optional `executor: NavigationExecutor` parameter to `NavigationSetup` to allow for usage outside of `NavHost` for compose navigation.


kboyarshinov:
will be addressed later with updated codegen solution for activities

PR #622, opened 2023-12-04
replace NavController parameter with optional NavEventNavigator

gabrielittner:
It's now possible to optionally pass a `NavEventNavigator` to `NavHost`. This navigator can be used to control navigation from outside the `NavHost`, the practical use case would be navigating based on bottom navigation or navigation drawer interactions. `NavHost` will automatically take care of the set up process. 

The `NavHostController` has been removed again as creating it before the deep link handling code has run breaks deep linking. We're now creating it internally again where we can fully control it and call `rememberNavController` after the `DisposableEffect` for deep links.

On the codegen side we're now requiring a `NavEventNavigator` to be provided into the Activity component and we'll always pass this to the `NavHost`. This matches the behavior of the codegen for destinations which also always requires a navigator even if you don't need one.


navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
```diff
@@ -55,8 +56,8 @@ import java.io.Serializable
  * provide a default set of url patterns that should be matched by any [DeepLinkHandler] that
  * doesn't provide its own [DeepLinkHandler.prefixes].
  *
- * The [navController] can be passed in optionally when needed, for example when using the NavHost
- * with a bottom navigation element.
+ * If a [NavEventNavigator] is passed it will be automatically set up and can be used to
+ * navigate within the `NavHost.
```
kboyarshinov, line 60:
```suggestion
 * navigate within the `NavHost`.
```

navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
```diff
@@ -98,8 +99,8 @@ public fun NavHost(
...
```
kboyarshinov, line 103:
```suggestion
 * navigate within the `NavHost`.
```

navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
```diff
@@ -138,17 +139,20 @@ private fun InternalNavHost(
...
     deepLinkHandlers: Set<DeepLinkHandler> = emptySet(),
     deepLinkPrefixes: Set<DeepLinkHandler.Prefix> = emptySet(),
-    navController: NavHostController = rememberNavController(),
+    navEventNavigator: NavEventNavigator? = null,
     destinationChangedCallback: ((BaseRoute) -> Unit)? = null,
     transitionAnimations: NavHostTransitionAnimations = NavHostDefaults.transitionAnimations(),
 ) {
     val context = LocalContext.current
 
+    DisposableEffect(context, deepLinkHandlers, deepLinkPrefixes) {
```
kboyarshinov, line 148: I think it's worth to add a comment here

```suggestion
    // should be called before rememberNavController to fill intent data with deeplinks
    DisposableEffect(context, deepLinkHandlers, deepLinkPrefixes) {
```

navigation-experimental/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
```diff
@@ -36,6 +37,9 @@ import kotlinx.coroutines.flow.map
  * provide a default set of url patterns that should be matched by any [DeepLinkHandler] that
  * doesn't provide its own [DeepLinkHandler.prefixes].
  *
+ * If a [NavEventNavigator] is passed it will be automatically set up and can be used to
+ * navigate within the `NavHost.
```
kboyarshinov, line 41:
```suggestion
 * navigate within the `NavHost`.
```

kboyarshinov (approved):
LGTM 👍

PR #625, opened 2023-12-04
support Modifier parameter in codegen and experimental navigation

gabrielittner:
I've introduced `SimpleNavHost` as a typealias for the nav host function that codegen requires. That makes future changes easier (they're still breaking changes since an additional parameter needs to be passed).

On the experimental navigation side the `Modifier` is for now used for a `Box` that wraps the content. This will probably change once we look into animations and/or predictive back. The AndroidX `NavHost` is using it on `AnimatedContent` that wraps the screens.


kboyarshinov (approved):
:shipit:

PR #627, opened 2023-12-05
use ImmutableSet for NavHost parameters

### LINKED ISSUE #609: Use `ImmutableSet` as NavHost parameter
This way the compose compiler can infer stability for it. 

On the codegen side we should generate a `@Provide fun provideImmutableDestinations(desintations: Set<NavDestination>): ImmutableSet<NavDestination> = this` and the same for deeplink related sets as part of the Activity generation.

gabrielittner:
Closes #609 

Allow compose to infer the immutability of the destination and deeplink sets by using `ImmutableSet` instead. In the codegen we're generating a provides for each set that turns Dagger's set into an immutable one.

I've also updated the internal `visibleEntries` list of the experimental implementation to use `ImmutableList`.


kboyarshinov (approved):
:shipit:

PR #628, opened 2023-12-05
call Navigation.setViewNavController on compose container view

gabrielittner:
Adding this for backwards compatibity with some code in our app.


navigation-compose/src/main/kotlin/com/freeletics/khonshu/navigation/compose/NavHost.kt
```diff
@@ -155,6 +157,15 @@ private fun InternalNavHost(
     val customActivityNavigator = remember(context) { CustomActivityNavigator(context) }
     val navController: NavHostController = rememberNavController(overlayNavigator, customActivityNavigator)
 
+    val composeView = LocalView.current
```
kboyarshinov, line 160: I think it's fine to have it as is now, but theoretically it can be made toggleable either with a flag or custom modifier.

kboyarshinov (approved):
👍

PR #629, opened 2023-12-05
extract interfaces out of NavEventNavigator

gabrielittner:
Extracts `Navigator`, `ResultNavigator`, `ActivityResultNavigator` and `BackInterceptor` out of `NavEventNavigator`. With this the whole public API of `NavEventNavigator`, except for the multi nav event method, is represented by these interfaces. For now this gives us 2 things:
- `NavigationExecutor` also implements `Navigator` to mirror the parts of the API for which it consumes `NavEvent`
- `NavEventCollector` is also just an implementation of `Navigator` which means the type doesn't need to be publicly exposed anymore.

For the longer term future this will allow us to move the event based navigation to a state based approach. For example `NavEventNavigator` (or a replacement for it) could just delegate to a shared implementation of `Navigator` instead of sending events which are then passed to a `Navigator` like right now.

One side effect of this is that `<reified T : NavRoute> navigateBackTo(inclusive: Boolean = false)` is now an extension function.


kboyarshinov (approved):
🚀

PR #631, opened 2023-12-06
deprecate Fragment support

gabrielittner:
🎉 

I want to do one release with other changes before we actually remove it.


josirichter (approved):
:shipit: ❤️

PR #633, opened 2023-12-06
split ComponentProvider by usage

gabrielittner:
Moves the `component` method used for Activities and the `ViewModel` to a separate file to make it clearer that these are their own thing. The viewmodel is now also called `ActivityComponentViewModel` to show it's purpose. I've also renamed the file for the remainder to `DestinationComponentLookup` to show these are used for generated `NavDestination` code.


kboyarshinov (approved):
👍

PR #664, opened 2024-01-10
introduce ActivityComponentProvider

gabrielittner:
`ActivityComponentProvider` mostly matches what we already have with `ComponentProvider` for destinations. It isolates the creation/finding of the dagger component and makes it reusable. We are also providing the created `ActivityComponentProvider` as an internal composition local from the generated activity. This will enable the next step which is to change the destination codegen to use the provider for its parent lookup. The 2 advantages of this are that destination code doesn't need `Context` anymore and that the activity scope/component can be used as parent.

### DIFF
--- a/codegen/src/androidMain/kotlin/com/freeletics/khonshu/codegen/internal/ActivityComponentLookup.kt
+++ b/codegen/src/androidMain/kotlin/com/freeletics/khonshu/codegen/internal/ActivityComponentLookup.kt
@@ -1,33 +1,46 @@
 package com.freeletics.khonshu.codegen.internal
 
 import android.content.Context
-import android.os.Bundle
+import androidx.activity.ComponentActivity
 import androidx.compose.runtime.DisallowComposableCalls
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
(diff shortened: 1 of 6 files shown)


codegen/src/androidMain/kotlin/com/freeletics/khonshu/codegen/internal/ActivityComponentLookup.kt
```diff
@@ -1,33 +1,46 @@
...
-    val store = ViewModelProvider(
-        viewModelStoreOwner,
-        SavedStateViewModelFactory(),
-    )[ActivityComponentViewModel::class.java]
-    return store.getOrCreate(C::class) {
-        val parentComponent = context.findComponentByScope<P>(parentScope)
-        val savedStateHandle = store.savedStateHandle
-        factory(parentComponent, savedStateHandle, arguments ?: Bundle.EMPTY)
+    if (requestedScope != activityScope) {
+        return activity.findComponentByScope(requestedScope)
```
gabrielittner, line 35: This allows to use the provider to also look up other components higher up in the hierarchy, e.g. AppScope. That way the parent lookup doesn't need to know whether the parent scope is an Activity scope or not.

kboyarshinov (approved):
LGTM 👍

PR #666, opened 2024-01-15
use ActivityComponentProvider to lookup parent components

gabrielittner:
Builds on top of #664.

This updates the destination codegen and `DestinationComponentLookup` to use the new `ActivityComponentProvider`, which
- allows using the activity scope as parent scope
- removes the need for `Context` from the destination component lookup


kboyarshinov (approved):
🚀

PR #668, opened 2024-01-17
limit StateMachine collection to RESUMED lifecycle

gabrielittner:
Prevents the state machine from running in the background with unintentional side effects like causing navigation events to happen which in turn causes AndroidX navigation to crash.

The file was moved from common to Android since it now has the dependency.


codegen/src/androidMain/kotlin/com/freeletics/khonshu/codegen/internal/CollectAsState.kt
```diff
@@ -0,0 +1,20 @@
...
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.flowWithLifecycle
+import com.freeletics.khonshu.statemachine.StateMachine
+
+@Composable
+@InternalCodegenApi
+public fun <S : Any> StateMachine<S, *>.asComposeState(): State<S?> {
+    val lifecycleOwner = LocalLifecycleOwner.current
+    return produceState<S?>(initialValue = null, lifecycleOwner) {
+        state.flowWithLifecycle(lifecycleOwner.lifecycle, Lifecycle.State.RESUMED)
```
hoc081098, line 17: I think `state` should be added to key list of `produceState`

gabrielittner: [b46b6bb](https://github.com/freeletics/khonshu/pull/668/commits/b46b6bb2bb3eac9731dc540d6fcb39e61f3b9ced) 
I've added a `remember` for the state as well because we can't assume that the `StateMachine` will return the same `Flow` on each call even if it's effectively the same `Flow` (e.g. it might do `channel.receiveAsFlow()` on each call).

hoc081098: Nice, that makes sense 🙏

kboyarshinov (approved):
:shipit:

PR #676, opened 2024-01-23
chore(Stack): use `ImmutableListAdapter` instead of `toImmutableList` to avoid unnecessary copying

hoc081098:
Although the returned `ArrayList` is mutable, but we can not mutate it -> `ImmutableListAdapter ` can be used in this case


gabrielittner (approved):
Thanks

PR #680, opened 2024-01-29
use NavHostTransitionAnimations.noAnimations() in generated code

gabrielittner:
Unblocks us from adopting codegen internally. I went with this since there is no transition implementation for experimental navigation yet, once we have something we could for example also add this to `SimpleNavHost`. For now we'll just go with no default transitions in the generated code.


kboyarshinov (approved):
👍

PR #705, opened 2024-02-28
remove copy of Anvil compilation

gabrielittner:
I've copied it before to be able to use a newer version of Kotlin compile testing but Anvil was updated to that version a long time ago, so it's not needed anymore.


kboyarshinov (approved):
👍

PR #707, opened 2024-02-29
Codegen: make it possible to enabele experimental nav at runtime

gabrielittner:
Previously when setting `experimentalNavigation = true` in `@NavHostActivity` it would only generate the experimental code. Since you can have the annotation twice it was already possible to use both at the same time, however it doesn't work well for launcher activities and activities launched through implicit intents, where you can't easily dynamically switch which is launched.

With this enabling `experimentalNavigation` will generate both the code for experimental and androidx navigation. The boolean to decide which one to use at runtime is injected from the Activity component. This allows providing a value based on a feature flag.

### DIFF
--- a/sample/simple/settings.gradle.kts
+++ b/sample/simple/settings.gradle.kts
@@ -13,5 +13,5 @@ plugins {
 rootProject.name = "simple-sample"
 
 configure<com.freeletics.gradle.plugin.SettingsExtension> {
-    includeKhonshu("../..", true)
+    includeKhonshu("../..")
 }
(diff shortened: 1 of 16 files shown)


sample/simple/settings.gradle.kts
```diff
@@ -13,5 +13,5 @@ plugins {
 rootProject.name = "simple-sample"
 
 configure<com.freeletics.gradle.plugin.SettingsExtension> {
-    includeKhonshu("../..", true)
+    includeKhonshu("../..")
```
gabrielittner, line 16: When `true` is passed the `includedBuild` will completely replace `navigation-compose` dependencies with `navigation-experimental` while using `false` (the default) will replace each of them with the matching local module.

gabrielittner: With this change it probably makes sense to remove the boolean option in FGP

kboyarshinov (approved):
👍

PR #708, opened 2024-02-29
update Anvil to 2.5.0-beta03 and use new GeneratedFileWithSources API

gabrielittner:
The Anvil update brings incremental compilation fixes. For those to work with Khonshu we need to use `GeneratedFileWithSources` instead of `GeneratedFile` in our `CodeGenerator`.


kboyarshinov (approved):
👍

PR #716, opened 2024-03-08
fix(navigation-experimental): fix an issue where `SaveableCloseable` can reference to wrong `SaveableStateHolder`

hoc081098:
- e.g. due to configuration changes
- because the lifetime of `SaveableCloseable` is longer than `SaveableStateHolder`


gabrielittner (approved):
Thanks for catching this and fixing it

PR #750, opened 2024-04-29
make codegen always use 'experimental' navigation

gabrielittner:
In preparation of removing AndroidX navigation the codegen will now always use Khonshu's own navigation implementation. The generated code will be as if `experimentalNavigation` is `true` (the option has been removed), but without the ability to decide between androidx and experimental at runtime.


codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/codegen/ActivityGenerator.kt
```diff
@@ -49,48 +46,18 @@ internal class ActivityGenerator(
...
-    private fun callNavHost(experimental: Boolean): CodeBlock {
-        return CodeBlock.builder()
-            .apply {
-                if (experimental) {
-                    addStatement("%M(", experimentalNavHost)
-                } else {
-                    addStatement("%M(", androidxNavHost)
-                }
-            }
+            .addStatement("%M(", experimentalNavHost)
```
befrvnk, line 49:
```suggestion
            .addStatement("%M(", navHost)
```

What do you say about renaming this to `navHost` to prevent confusion?

PR #769, opened 2024-05-07
expose current NavRoot in destination change callback

gabrielittner:
Primary use case for this is updating the selected state of something like bottom tabs or a navigation rail. When switching between destinations that are within a stack.

### DIFF
--- a/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
+++ b/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
@@ -139,7 +139,7 @@ private fun DestinationChangedCallback(
     if (destinationChangedCallback != null) {
         val root = snapshot.root
         val current = snapshot.current
-        DisposableEffect(root, current) {
+        DisposableEffect(destinationChangedCallback, root, current) {
             destinationChangedCallback(root.route as NavRoot, current.route)
             onDispose {}
         }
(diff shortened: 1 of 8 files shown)


navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
```diff
@@ -134,12 +134,13 @@ private fun SystemBackHandling(snapshot: StackSnapshot, executor: MultiStackNavi
...
     snapshot: StackSnapshot,
-    destinationChangedCallback: ((BaseRoute) -> Unit)?,
+    destinationChangedCallback: ((NavRoot, BaseRoute) -> Unit)?,
 ) {
     if (destinationChangedCallback != null) {
+        val root = snapshot.root
         val current = snapshot.current
-        DisposableEffect(current) {
-            destinationChangedCallback(current.route)
+        DisposableEffect(root, current) {
```
hoc081098, line 142: I think `destinationChangedCallback` should be added to the key list, or we use `rememberUpdatedState` instead 🙏

gabrielittner: Thanks for catching this [e8d2abc](https://github.com/freeletics/khonshu/pull/769/commits/e8d2abc3e149a083704d0733296c2bbacb295b23)

PR #779, opened 2024-05-15
replace onEntryRemoved callback with closing entries

gabrielittner:
This makes it a little bit more direct. We can't avoid referencing the store in 2 places: we want it in the entry because that's how we access it but the the `ViewModel` needs to hold it as well so that it stays alive during configuration changes. Instead of passing down the `onEntryRemoved` callback everywhere, the store will now remove itself from the `ViewModel` when it's closed and we close the store through the entry on removal of the entry.


kboyarshinov (approved):
LGTM 👍

PR #780, opened 2024-05-15
replace NavigationExecutor with StackSnapshot for back stack access

gabrielittner:
Part 1 of 2 for starting to use `StackSnapshot` for accessing things like the route, store and handle. This replaces the methods on `NavigationExecutor` to get these and always goes through the `StackSnapshot`. 

The missing part is to use the `StackSnapshot` and `StackEntry` that we already have inside `NavHost` directly instead of relying on `NavigationExecutor` to get an instance. Only with this step the issue that the backstack already changed but compose didn't update yet that can lead to crashes will be fixed.


kboyarshinov (approved):
Nice 👏

PR #781, opened 2024-05-15
pass StackSnapshot to content composable

gabrielittner:
Part 2 for #780. The internal `*Destination` functions that have the `extra: Any?` for component providers will now get the `StackEntry` instead of the `BaseRoute` plus the `StackSnapshot` passed to the `Composable` lambda. This allows the codegen to then do no lookups for the current entry and use the snapshot for looking up the parent if needed. 

For the public API nothing changes, the `BaseRoute` keeps being the parameter passed to the composable. Once all of these are a bit more mature API wise we can also think about exposing the entry/snapshot/store publicly.


kboyarshinov (approved):
👍

PR #798, opened 2024-05-28
avoid ConcurrentModificationException when clearing store

gabrielittner:
We're seeing a `ConcurrentModificationException` in this code. So far it only happened in Google Play pre launch report and I assume it's an edge case when a machine interacts with the app (extremely fast interactions). Since this is happening when the ViewModel is cleared and the Activity is being destroyed I just made the clearing more defensive, instead of making everything in the ViewModel thread safe.


kboyarshinov (approved):
👍

PR #801, opened 2024-05-28
extract Activity result and event logic

gabrielittner:
Background info:
- With the introduction of `HostNavigator` regular navigation does not require nav events anymore
- Activity results will always need them because we need to create the launchers and we wouldn't want to do that globally for any possible contract that an app needs.
    - A small exception could be the permission logic, we could have the `PermissionResultContract` in the `HostNavigator` and create a launcher in the `Activity` and then don't need it separately in each destination that needs permission requests. However even that would still require nav events.

Based on that this exctracts the event logic and required code for activity results to `ActivityResultNavigator`. `NavEventNavigator` extends that so from a consumer perspective nothing really changes. This will then allow having a destination level `Navigator` class that doesn't use nav events for anything else but instead delegates to `HostNavigator` for regular navigation.

### DIFF
--- a/navigation/api/android/navigation.api
+++ b/navigation/api/android/navigation.api
@@ -5,11 +5,15 @@ public final class com/freeletics/khonshu/navigation/ActivityDestination : com/f
 	public final fun getIntent ()Landroid/content/Intent;
 }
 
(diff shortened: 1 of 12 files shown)


kboyarshinov (approved):
:+1:

PR #804, opened 2024-05-29
make HostNavigator implement ResultNavigator

gabrielittner:
With this it's not needed anymore to use nav events for local navigation results. The implementation is very simple at it's core since we have direct access to the `SavedStateHandle` on both sides and can simply read/write from it. So the `ResultOwner` that is returned by the register can simply hold the `SavedStateHandle` and the result flow is just a mapped version of `SavedStateHandle.getStateFlow`. 

However for the time where we still have `NavEventNavigator` the `ResultOwner` structure becomes a bit more complicated. `NavigationResultRequest` is now a base class with one implementation for `HostNavigator` (the one described above) and one for nav events. Once we remove `NavEventNavigator` we can remove the latter and merge the former back into the base class. The intermediate `ManagedResultOwner` can then also be merged into `ContractResultOwner`.


kboyarshinov (approved):
LGTM 👍

PR #805, opened 2024-05-29
Codegen: replace NavEventNavigator with ActivityResultNavigator

gabrielittner:
This allows using other subclasses of `ActivityResultNavigator` as your navigator instead of just `NavEventNavigator`. On the consumer side the only change is that `@ContributesBinding(..., NavEventNavigator::class)` becomes `@ContributesBinding(..., ActivityResultNavigator::class)`.


sample/simple/gradle.properties
```diff
@@ -6,3 +6,5 @@ android.useAndroidX=true
 kapt.include.compile.classpath=false
 
 fgp.android.namespacePrefix=com.freeletics
+fgp.kotlin.anvilKsp=true
+fgp.kotlin.anvilKspWithComponent=false
```
gabrielittner, line 10: Missed this during the k2 update.

sample/simple/feature/main/src/main/AndroidManifest.xml
```diff
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
     <application>
         <activity
-            android:name="com.freeletics.khonshu.sample.feature.main.KhonshuExperimentalMainScreenActivity"
+            android:name="com.freeletics.khonshu.sample.feature.main.KhonshuMainScreenActivity"
```
gabrielittner, line 5: Missed this during the AndroidX cleanup

kboyarshinov (approved):
👍

PR #809, opened 2024-06-05
make HostNavigator impelement BackInterceptor

gabrielittner:
This is the last piece of functionality that is made independent of nav events. With this it's possible to directly use `HostNavigator` for intercepting back.

### DIFF
--- a/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
+++ b/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
@@ -111,7 +111,7 @@ internal class SaveableCloseable(
 }
 
 @Composable
-private fun SystemBackHandling(snapshot: StackSnapshot, navigator: Navigator) {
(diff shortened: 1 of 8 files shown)


navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
```diff
@@ -131,9 +131,11 @@ private fun SystemBackHandling(snapshot: StackSnapshot, navigator: Navigator) {
 
     DisposableEffect(backPressedDispatcher, callback) {
         backPressedDispatcher.onBackPressedDispatcher.addCallback(callback)
+        backPressedDispatcher.onBackPressedDispatcher.addCallback(navigator.onBackPressedCallback)
```
gabrielittner, line 134: Theoretically we could combine these 2 but we'd need to split them again once we support predictive back (since that needs to be handled inside nav host) so I kept the existing code as is.

kboyarshinov (approved):
👍

PR #810, opened 2024-06-05
introduce DestinationNavigator as replacement for NavEventNavigator

gabrielittner:
Similar to `NavEventNavigator`, `DestinationNavigator` is meant to be used by individual screens (destinations) for their navigation actions. The difference is that most actions are delegated to `HostNavigator`, but it also provides activity result contract functionality which is still and will continue to be event based. 

Why have a base class?
- Internally it worked well for us to have a navigator class in each screen to isolate navigation logic. While you could do the delegation yourself and extend `ActivityResultNavigator` where needed this provides some consistency. 
- There is no additional step needed in case you need activity results or some other functionality.
- In the codegen we can simply expect each screen to provide an `ActivityResultNavigator`.

Since `HostNavigator` is provided in the `ActivityScope` the default parent scope for `@NavDestination` has been changed to `ActivityScope` to make this simpler.


kboyarshinov (approved):
🚀

PR #816, opened 2024-06-11
add navigate {} to DestinationNavigator

gabrielittner:
Follow up to #803, since `navigate {}` is available in `HostNavigator` now, we also want it in `DestinationNavigator`. In this case we need to manually delegate since it's not part of the `Navigator` interface (we could think about adding it to that, the only downside is that it would allow nested calls to it).


kboyarshinov (approved):
👍

PR #819, opened 2024-06-11
add TestHostNavigator

gabrielittner:
Adds a test helper for using `HostNavigator` which makes interactions with both `HostNavigator` and `DestinationNavigator` testable. This initial implementation will use nav events internally so that the existing test code  can be re-used. Once we remove nav events we can re-think this.

I've also added a `Boolean` return value to `handleDeepLink` this allows us to do other navigation actions based on whether the method did something or not.

### DIFF
--- a/navigation-testing/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavigatorTurbine.kt
+++ b/navigation-testing/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavigatorTurbine.kt
@@ -30,7 +30,7 @@ public suspend fun TestHostNavigator.test(
 }
 
 /**
- * Collects events from [DestinationNavigator] and and allows the [validate] lambda to consume
+ * Collects events from [DestinationNavigator] and allows the [validate] lambda to consume
  * and assert properties on them in order. If any exception occurs during validation the
  * exception is rethrown from this method.
  *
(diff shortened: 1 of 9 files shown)


kboyarshinov (approved):

navigation-testing/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavigatorTurbine.kt
```diff
@@ -9,6 +9,105 @@ import com.google.common.truth.Truth
 import kotlin.reflect.KClass
 import kotlin.time.Duration
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.merge
+
+/**
+ * Collects events from [TestHostNavigator] and and allows the [validate] lambda to consume
```
kboyarshinov, line 15:
```suggestion
 * Collects events from [TestHostNavigator] and allows the [validate] lambda to consume
```

navigation-testing/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavigatorTurbine.kt
```diff
@@ -9,6 +9,105 @@ import com.google.common.truth.Truth
...
+    validate: suspend NavigatorTurbine.() -> Unit,
+) {
+    navEventNavigator.navEvents.test(timeout, name) {
+        val turbine = DefaultNavigatorTurbine(navEventNavigator, this)
+        validate(turbine)
+    }
+}
+
+/**
+ * Collects events from [DestinationNavigator] and and allows the [validate] lambda to consume
```
kboyarshinov, line 33:
```suggestion
 * Collects events from [DestinationNavigator] and allows the [validate] lambda to consume
```

kboyarshinov (approved):
👍

PR #820, opened 2024-06-12
fix 2 issues with replaceAll

gabrielittner:
1. Existing issue: `val startRoot` was assigned once during the creation of `MultiStack` which means the value is wrong after `replaceAll` changes the `startStack`. The impact is negligible so far since it's only used for 
2. Unreleased issue: Currently when `MultiStack` is created from saved state the navigator if the `startRoot` passed to `NavHost`/`rememberHostNavigator` does not match the saved `NavRoot` we discard the state. This makes sense since from a compose perspective changing the given `startRoot` should re-compose everything and you effectively get a new `NavHost`. However if we create a `HostNavigator` through outside of compose (like we do in the codegen) this behavior doesn't make sense. For example we create the `HostNavigator` with a dummy startRoot and then always use `replaceAll` on the initial start logic and with that we always want to keep the saved state. The solution keeps the behavior for compose but disables it for codegen.


kboyarshinov (approved):
👍

PR #827, opened 2024-06-19
update docs for HostNavigator related changes

gabrielittner:
Updates the docs with all the recent changes
- introduction of HostNavigator
- removal of AndroidX navigation
- the not so recent deep link gradle plugin


docs/navigation/activities.md
```diff
@@ -1,58 +1,70 @@
...
+## Setup
+
+Navigating to an `Activity` is not directly possible with `HostNavigator`. Instead an
+`ActivityNavigator` is needed. This can just be a simple class that extends `ActivityNavigator`
+or `DestinationNavigator`. The latter combines the functionality of `HostNavigator` and
+`ActivityNavigator` so that one class can be used for all navigation related actions of a
+screen.
+
+From the `Composable` it is then required to call `NavigationSetup(navigator)` which
+will do the required setup to make the `Activity` related navigation actions work.
```
kboyarshinov, line 17: Since this is done by codegen, it is probably not needed to be mentioned here, unless codegen isn't used. What do you think?

docs/navigation/activities.md
```diff
@@ -1,58 +1,70 @@
...
 ```
 
-`SettingsActivity` can obtain an instance of the route that was used to navigate to it by calling
-`Activity.getRoute()` or `Activity.requireRoute()`.
+An instance of the route will automatically be added to the `Intent` which is why it's not needed to
+manually put the `id` into it. `SettingsActivity` can then obtain an instance of the route that was used
+to navigate to it by calling `Activity.getRoute()` or `Activity.requireRoute()` and access the
+argument through it.
+
+To avoid having to reference a specific `Activity` `class` it's possible to use customt `Intent` actions.
```
kboyarshinov, line 39: typo

```suggestion
To avoid having to reference a specific `Activity` `class` it's possible to use custom `Intent` actions.
```

docs/navigation/activities.md
```diff
@@ -61,31 +73,87 @@ class ShareRoute(
...
+
+## Activity results
 
-val browserDestination: NavDestination = ActivityDestination<BrowserRoute>(
-    // basic intent that is extended with the Intent above
-    intent = Intent(Intent.ACTION_VIEW)
-)
+External SDKs and the framework often provide an `Activity` that is supposed to be used with
+`startActivityForResult`. AndroidX already introduced `ActivityResultContract` to simplify handling
+this and `ActivityNavigator` uses them to also enable starting them from outside the UI layer
```
kboyarshinov, line 91: rephrased since it's not obvious what `them` is

```suggestion
this and `ActivityNavigator` uses it to also enable starting Activities from outside the UI layer
```

docs/navigation/activities.md
```diff
@@ -61,31 +73,87 @@ class ShareRoute(
...
+and receiving results there.
+
+To use the API `registerForActivityResult` needs to be called with an instance of the wanted
+`ActivityResultContract`. This needs to happen before `NavigationSetup` is called for the navigator,
+so it is recommended to do this during the construction of the navigator. The method returns an
```
kboyarshinov, line 96: > This needs to happen before `NavigationSetup` is called for the navigator,
so it is recommended to do this during the construction of the navigator. 

Isn't it the only way now? `registerForActivityResult` has to be done either in constructor or as `val` property of navigator. Anything else would be late and cause runtime crash.

docs/navigation/deeplinks.md
```diff
@@ -43,48 +43,111 @@ class UserProfileDeepLinkHandler : DeepLinkHandler {
...
 
+
+### Gradle plugin
+
 For the uri based deep link handling to work regular `intent-filter`s need to be created
-in the `AndroidManifest`.
+in the `AndroidManifest`. To automate this process there is a Gradle plugin that reads
+deep link definitions from a `toml` file and automatically generates the `Intent` filters
+in the `AndroidManifest`. There is then a test helper that allows writing a unit test
+verify that the `toml` file and the defined `DeepLinkHandler` classes handle the same
```
kboyarshinov, line 78:
```suggestion
in the `AndroidManifest`. There is then a test helper that allows writing a unit test
to verify that the `toml` file and the defined `DeepLinkHandler` classes handle the same
```

docs/navigation/get-started.md
```diff
@@ -1,50 +1,30 @@
...
-   feature module it's not usable by other features either. That meant that in the end we didn't get
-   the nice type safety advantages of safe-args and the XML was just boilerplate that didn't
-   form a real graph in the end.
-
+We believe that navigation should be triggered from the business logic. This way it is
+separated from the UI layer and becomes easily testable in isolation without having to
+actually navigate inside a running app.
+
+Originally this library was a wrapper around AndroidX navigation which was eventually
+removed from the internals.The initial XML based definitions of destinations
```
kboyarshinov, line 8:
```suggestion
removed from the internals. The initial XML based definitions of destinations
```

docs/navigation/results.md
```diff
@@ -97,31 +15,36 @@ from the `key` property of the request object.
 The target screen can then simply call `devliverNavigationResult(route.key, result)` on its own
```
kboyarshinov, line 15: typo in `deliverNavigationResult`

kboyarshinov (approved):
👍

docs/navigation/activities.md
gabrielittner: I mostly mentioned it to keep the navigation docs codegen free and describe how it would work when doing it manually. Similar to how it's also talking about how to define the destinations

docs/navigation/activities.md
gabrielittner: I think there are some ways where you could do the registration between constructor and setup but let's change it:
```suggestion
so it needs to be called during the construction of the navigator. The method returns an
```

docs/navigation/results.md
gabrielittner: [e3d5a22](https://github.com/freeletics/khonshu/pull/827/commits/e3d5a221ef94283636d8b24c6e8fe2c32b004a52)

PR #843, opened 2024-07-12
support KSP2 in codegen

gabrielittner:
The tests are now also run with ksp2 in addition to ksp1. There is a small behavior difference in how lambdas are represented in types which I added a fix for.


codegen-compiler/src/main/kotlin/com/freeletics/khonshu/codegen/codegen/ComponentComposableGenerator.kt
```diff
@@ -55,7 +55,7 @@ internal class ComponentComposableGenerator(
                     addStatement("val scope = %M()", rememberCoroutineScope)
                         .beginControlFlow(
                             "val sendAction: %T = %M(stateMachine, scope)",
-                            data.sendActionParameter!!.typeName.functionToLambda(),
+                            data.sendActionParameter!!.typeName,
```
gabrielittner, line 58: The `functionToLambda` call now always happens during parsing.

zaream (approved):

PR #846, opened 2024-07-16
prepare for predictive back support

gabrielittner:
This adds an internal `previous: StackEntry<*>` to the snapshot so that we can show the previous screen for predictive back support. For proper multi back stack support this means that we need the `StackEntry` of the start stack inside the snapshot so that we are able to show that as well.

### DIFF
--- a/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/internal/StackSnapshot.kt
+++ b/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/internal/StackSnapshot.kt
@@ -23,7 +23,7 @@ public class StackSnapshot internal constructor(
         get() = entries.last()
 
     internal val previous: StackEntry<*>
-        get() = entries.getOrNull(entries.size - 2) ?: startStackRootEntry
+        get() = entries.getOrNull(entries.lastIndex - 1) ?: startStackRootEntry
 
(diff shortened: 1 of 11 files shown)


navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/internal/StackSnapshot.kt
```diff
@@ -22,8 +22,11 @@ public class StackSnapshot internal constructor(
     internal val current: StackEntry<*>
         get() = entries.last()
 
+    internal val previous: StackEntry<*>
+        get() = entries.getOrNull(entries.size - 2) ?: startStackRootEntry
```
gabrielittner, line 26: If `current` is `startStackRootEntry` then what is returned here is technically wrong, but in practice `previous` is never used in that case.

kboyarshinov, line 26: What do you think of using `lastIndex - 1` instead of `size - 2` to get index before the last?

```suggestion
        get() = entries.getOrNull(entries.lastIndex - 1) ?: startStackRootEntry
```

kboyarshinov (approved):
LGTM 👍

PR #847, opened 2024-07-16
implement predictive back support

gabrielittner:
Initial support for predictive back that is based on [this spec](https://developer.android.com/design/ui/mobile/guides/patterns/predictive-back#full-screen-surfaces) (see comments in code for some more details). I didn't add an option for this since it can be controlled through `android:enableOnBackInvokedCallback="true"` which defaults to off even on Android 15.

### DIFF
--- a/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
+++ b/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
@@ -1,17 +1,23 @@
 package com.freeletics.khonshu.navigation
 
-import androidx.activity.OnBackPressedCallback
+import android.view.animation.PathInterpolator
 import androidx.activity.compose.LocalOnBackPressedDispatcherOwner
+import androidx.activity.compose.PredictiveBackHandler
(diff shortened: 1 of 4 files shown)


navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
```diff
@@ -110,45 +129,106 @@ internal class SaveableCloseable(
...
+ * Animation for transitioning out the screen currently on top of the back stack. The specs are
+ * following the "Full screen surfaces" section of the
+ * [predictive back design guide](https://developer.android.com/design/ui/mobile/guides/patterns/predictive-back#full-screen-surfaces)
+ * with 2 differences:
+ *
+ * - Horizontal translation was added for a better feeling.
+ * - The spec uses the 35% mark as the transition point between the screens where none of them is visible, while this
+ *   implementation uses 3.5% (0.035f). With the original 35% value it was easily possible to have situations where
+ *   the current screen is still shown but stopping the gesture would already navigate back, which results in a very
+ *   weird user experience.
```
gabrielittner, line 142: This will probably need some more tweaks in the future

PR #850, opened 2024-07-18
fix race condition with 2 DisposableEffect for back handlers

gabrielittner:
`PredictiveBackHandler` uses a `DisposableEffect` to add its `OnBackPressedCallback` to the dispatcher and then we do the same for the `HostNavigator`'s callback. The order is important because if `navigator.backClicks` is collected it needs to take precedence over the predictive back handler. While testing I found some cases where this is not the case which resulted in custom back handling not working.

At least as an initial fix this copies `PredictiveBackHandler` and adds a parameter for an additional callback to it. This way both callbacks are added in the `DisposableEffect` and we can have our ordering guarantee.

### DIFF
--- a/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
+++ b/navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
@@ -1,8 +1,6 @@
 package com.freeletics.khonshu.navigation
 
 import android.view.animation.PathInterpolator
-import androidx.activity.compose.LocalOnBackPressedDispatcherOwner
-import androidx.activity.compose.PredictiveBackHandler
 import androidx.compose.animation.core.Animatable
 import androidx.compose.foundation.layout.Box
 import androidx.compose.runtime.Composable
@@ -19,6 +17,7 @@ import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.util.lerp
 import com.freeletics.khonshu.navigation.deeplinks.DeepLinkHandler
+import com.freeletics.khonshu.navigation.internal.PredictiveBackHandler
 import com.freeletics.khonshu.navigation.internal.StackEntry
 import com.freeletics.khonshu.navigation.internal.StackSnapshot
 import java.io.Closeable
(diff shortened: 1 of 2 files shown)


navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
```diff
@@ -203,36 +202,45 @@ private fun systemBackHandling(snapshot: StackSnapshot, navigator: HostNavigator
...
+    } catch (e: IllegalStateException) {
+        // The exception is thrown when navigateBack is called while the backstack is at the root. If the
+        // system back is triggered twice very quickly after each other there is a short time window
+        // after the first where the OnBackPressedCallback is not yet updated and would then also handle the
+        // second. This suppresses the crash in that case.
+        if (snapshot.value.canNavigateBack) {
+            throw e
         }
     }
+}
```
gabrielittner, line 235: Brought back the try/catch around `navigateBack` the removal wasn't intentional when implementing predictive back.

navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
```diff
@@ -203,36 +202,45 @@ private fun systemBackHandling(snapshot: StackSnapshot, navigator: HostNavigator
...
 
-    return backProgress.asState()
+private suspend fun Animatable<Float, *>.tryAnimateTo(value: Float) {
+    try {
+        animateTo(value)
+    } finally {
+        // make sure that the animation is not stuck at intermediate value in case animateTo gets cancelled
+        snapTo(value)
+    }
 }
```
gabrielittner, line 244: This is better than the final value approach because even if `animateTo` is cancelled we should still call `navigateBack`.

navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavHost.kt
```diff
@@ -203,36 +202,45 @@ private fun systemBackHandling(snapshot: StackSnapshot, navigator: HostNavigator
...
 
-    return backProgress.asState()
+private suspend fun Animatable<Float, *>.tryAnimateTo(value: Float) {
+    try {
+        animateTo(value)
+    } catch (e: CancellationException) {
+        // make sure that the animation is not stuck at intermediate value in case animateTo gets cancelled
+        snapTo(value)
+    }
 }
```
gabrielittner, line 244: This is better than the `finalValue` approach because even if animateTo is cancelled we should still call navigateBack.

kboyarshinov (approved):
👍

PR #854, opened 2024-07-22
use movableContentOf to avoid blinking

gabrielittner:
Currently when the animation finished and the previous entry becomes the current entry, the screen blinks once at the moment when the re-composition happens. This is because the composable is removed at it's previous place and added again in the new place. With this we're using `movableContentOf` to avoid that blinking and letting compose to re-use the already added composable.

https://github.com/user-attachments/assets/24e36e37-fa20-4cb9-acdd-c36026360056 

https://github.com/user-attachments/assets/c4c8ab1d-0c2f-4b82-bccf-148bc6d86c0e


kboyarshinov (approved):
LGTM 👍

PR #867, opened 2024-08-05
also consider root from start stack in entryFor

gabrielittner:
With predictive back the root of the start stack might be accessed through this even when we're in a different stack.


kboyarshinov (approved):
👍

PR #922, opened 2024-10-24
refactor NavRoot/backstack related navigation methods

gabrielittner:
To reduce confusion and the potential for mistakes/crashes this changes the naming and behavior of `NavRoot` navigation methods.

**Before**
* `navigateToRoot(root: NavRoot, restoreRootState: Boolean = false)`
    * if no back stack for `root` exists, a new one is created and will be made the current back stack
    * if a back stack for `root` exists, it will be made the current back stack
        * if `restoreRootState` is `false` the found existing back stack will be cleared so that it only contains `root`
        * if `restoreRootState` is `true` the found existing back stack will be shown as is (this is what the bottom nav uses when switching tabs)
    * calling this with the current back stack will result in a crash since we’re already in the right back stack
* `resetToRoot(root: NavRoot)`
    * clears the back stack so that `root` is shown
    * `root` needs to be the root of the current back stack or the start back stack
    * calling this with a `root` that doesn’t match either will result in a crash
* `replaceAll(root: NavRoot)`
    * removes all back stacks and creates a new one with `root`
    * used when switching between logged in and logged out

**New**
* `switchBackStack(root: NavRoot)`
    * equivalent to `navigateToRoot(root, restoreRootState = true)`
* `showRoot(root: NavRoot)`
    * equivalent to `resetToRoot(root)` when `root` is the current back stack at it’s root
    * equivalent to `navigateToRoot(root, restoreRootState = false)` when `root` is not the current back stack
* `replaceAllBackStacks(root: NavRoot)`
    * equivalent to `replaceAll(root)`

It's the same amount of methods but they are focusing more on the outcome, previously the 2 cases that `showRoot` was covering effectively both “I want to see root” but you needed to choose the right method based on the current state. This could lead to crashes or unneeded complexity, especially for screens that could be used in multiple stacks. Compared to before none of these crash for any combination of current back stack and given `root`.

### DIFF
--- a/navigation-testing/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavigatorTurbine.kt
+++ b/navigation-testing/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavigatorTurbine.kt
@@ -129,16 +129,6 @@ public interface NavigatorTurbine {
     public suspend fun awaitNavigateTo(route: NavRoute)
 
     /**
-     * Assert that the next event received was a navigation event to the given [root]. This function
-     * will suspend if no events have been received.
-     *
-     * @throws AssertionError - if the next event was not a matching event.
-     */
-    public suspend fun awaitSwitchBackStack(
-        root: NavRoot,
-    )
-
-    /**
      * Assert that the next event received was a navigation event to the given [route]. This function
      * will suspend if no events have been received.
      *
@@ -182,24 +172,28 @@ public interface NavigatorTurbine {
     )
 
     /**
-     * Assert that the next event received was a "reset to root" navigation event with matching
+     * Assert that the next event received was a "switch back stack" navigation event with matching
      * parameters. This function will suspend if no events have been received.
      *
      * @throws AssertionError - if the next event was not a matching event.
      */
-    public suspend fun awaitShowRoot(
-        root: NavRoot,
-    )
+    public suspend fun awaitSwitchBackStack(root: NavRoot)
 
     /**
-     * Assert that the next event received was a "replace all" navigation event with matching
+     * Assert that the next event received was a "show root" navigation event with matching
      * parameters. This function will suspend if no events have been received.
      *
      * @throws AssertionError - if the next event was not a matching event.
      */
-    public suspend fun awaitReplaceAllBackStacks(
-        root: NavRoot,
-    )
+    public suspend fun awaitShowRoot(root: NavRoot)
+
+    /**
+     * Assert that the next event received was a "replace all back stacks" navigation event with matching
+     * parameters. This function will suspend if no events have been received.
+     *
+     * @throws AssertionError - if the next event was not a matching event.
+     */
+    public suspend fun awaitReplaceAllBackStacks(root: NavRoot)
 
     /**
      * Assert that the next event received was a navigate for result event to [request].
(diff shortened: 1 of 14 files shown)


navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/Navigator.kt
```diff
@@ -39,20 +29,30 @@ public interface Navigator {
...
+     * The back stack that is shown at the time of calling this method will not back stack will not be modified.
+     */
+    public fun switchBackStack(root: NavRoot)
+
+    /**
+     * Show [root] as the current destination. If there already is a back stack for [root], regardless of whether
+     * it is the current back stack or not, the stack will be cleared.
+     *
+     * The back stack that is shown at the time of calling this method, if it isn't the back stack of [root],
+     * will not back stack will not be modified.
```
gabrielittner, line 45: I'm not 100% sure about this part. It matches `resetToRoot(root)` when `root` is the current stack and it also matches `navigateToRoot(root, restoreRootState = false)` when `root` isn't current. However `resetToRoot(root)` where `root` is not the current stack but the start stack would clear both current and the start stack before. If we do change this to also clear the current stack, it would not be completely equivalent to `navigateToRoot(root, restoreRootState = false)` for the case that `root` is the start destination.

For how we use the existing methods in our code base, clearing the current stack would be more fitting. We do have multiple cases where we call `resetToRoot(StartRoot)` and expect the current stack to be cleared (usually at the end of a flow). We have 0 cases of `navigateToRoot(root, restoreRootState = false)` so it not being entirely equivalent doesn't matter to existing code.

Despite that, I think the way it's implemented in this PR and written in the doc here is better and we should probably just update the cases that we have.

navigation-testing/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavigatorTurbine.kt
```diff
@@ -188,7 +187,7 @@ public interface NavigatorTurbine {
      *
```
josirichter, line 187: nit: description in ln 185 needs to be adapted too

navigation-testing/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavigatorTurbine.kt
```diff
@@ -198,7 +197,7 @@ public interface NavigatorTurbine {
...
```
josirichter, line 197: nit: description in ln 195 needs to be adapted too

navigation-testing/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavigatorTurbine.kt
gabrielittner: https://github.com/freeletics/khonshu/pull/922/commits/4b3383c5db9595f230ae8781bf2a71f2cefa2001

navigation-testing/src/androidMain/kotlin/com/freeletics/khonshu/navigation/NavigatorTurbine.kt
gabrielittner: https://github.com/freeletics/khonshu/pull/922/commits/4b3383c5db9595f230ae8781bf2a71f2cefa2001

PR #932, opened 2024-11-05
remove Anvil plugin

gabrielittner:
This will require using KSP for Khonshu codegen. Removing the old plugin allows us to drop Kotlin 1.9 which in turn will enable some other improvements like using `kotlin.AutoCloseable` instead of `java.io.Closeable`.


kboyarshinov (approved):
👍

PR #936, opened 2024-11-11
fix awaitNavigate

gabrielittner:
Turbine's `takeEvent()` has a check that's it's not called from a suspending function. We don't do that (`fun navigate(block: Navigator.() -> Unit)` is not suspending), however we can still trigger the check when `navigate` is called from a suspending function. That's because `takeEvent` just checks whether `invokeSuspend` appears in the stacktrace. To avoid this issue we're not reading the events from the internal channel.


kboyarshinov (approved):
👍

PR #1098, opened 2025-07-16
use AutoCloseable instead of Java's Closeable

gabrielittner:
No functionality change, just a kmp compatible type instead of a jvm specific one.


kboyarshinov (approved):
🚢

PR #1105, opened 2025-07-23
integrate FlowReduxStateMachineFactory

gabrielittner:
Allows using FlowReduxStateMachineFactory with the `produceStateMachine` function. Will wait for cutting a first release of flowredux2 before merging so that we don't use a snapshot.


kboyarshinov (approved):
👍

PR #1192, opened 2025-11-05
bind DestinationNavigator instead of ActivityNavigator

gabrielittner:
Use `DestinationNavigator` in codegen since we're more opinionated there and it's the main implementation of the interface.


kboyarshinov (approved):
👍

PR #1193, opened 2025-11-06
split MultiStack creation from HostNavigator creation

gabrielittner:
Just an internal change to separate the code a bit more and make some following changes around saving state and view models easier.


kboyarshinov (approved):
👍

PR #1203, opened 2025-11-10
use Compose saveable APIs for saving state

gabrielittner:
Migrates from just using bundles which won't work for KMP to the Comose Saveable APIs.


kboyarshinov (approved):
LGTM 👍

PR #1208, opened 2025-11-12
use Comopose retain API instead of ViewModel

gabrielittner:
Uses the new retain API to keep the StackEntryStores and through that the Metro graphs during configuration changes. For now we still keep the class that holds them, it's just not a view model anymore. We might be able to remove this in the future by having custom retain scopes that are tied to the stack entries, but that can be a separate change (also not 100% sure about this working because we need to access parent stores).


kboyarshinov (approved):
👍

PR #1371, opened 2026-04-12
support codegen for iOS

gabrielittner:
Similar to the desktop codegen this adds an equivalent of the Activtity codegen to generate an entry point for using codegen with a nav host in an iOS app. Other than the generation of the different class there aren't really changes and even that class is mostly the same.


kboyarshinov (approved):
🚀

PR #1372, opened 2026-04-12
add iOS target to sample

gabrielittner:
Needs #1371 to be merged first. Because FGP does include the x64 target by default I had to enable it for the libraries as well and downgrade androidx annotations since it just dropped that.


sample/simple/app/simple/src/iosMain/kotlin/com/freeletics/simple/Controller.kt
```diff
@@ -0,0 +1,5 @@
+package com.freeletics.simple
+
+import platform.UIKit.UIViewController
+
+expect fun createViewController(): UIViewController
```
gabrielittner, line 5: I've added this as a simple wrapper for the first iteration to have this easily available. The expect/actual is needed because ksp generates sources in the concrete iOS targets so you don't see it from `iosMain`.

kboyarshinov (approved):
🚀

PR #1378, opened 2026-04-21
Fallback to default annotation values when parsing

kboyarshinov:
This would allow apps not have to redeclare annotations with defaults in KMP non JVM targets

### DIFF
--- a/codegen-compiler/src/jvmMain/kotlin/com/freeletics/khonshu/codegen/parser/KspParser.kt
+++ b/codegen-compiler/src/jvmMain/kotlin/com/freeletics/khonshu/codegen/parser/KspParser.kt
@@ -7,6 +7,7 @@ import com.freeletics.khonshu.codegen.HostViewControllerData
 import com.freeletics.khonshu.codegen.HostWindowData
 import com.freeletics.khonshu.codegen.Navigation
 import com.freeletics.khonshu.codegen.util.activityScope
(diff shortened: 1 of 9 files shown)


codegen-compiler/src/jvmMain/kotlin/com/freeletics/khonshu/codegen/parser/KspParser.kt
```diff
@@ -225,7 +251,7 @@ private fun KSAnnotation.stateMachineParameters(logger: KSPLogger): Triple<Class
 }
 
 private fun KSAnnotation.parentScopeExtendsBaseRoute(): Boolean {
-    val declaration = parentScopeType.declaration as KSClassDeclaration
+    val declaration = (parentScopeType ?: return false).declaration as KSClassDeclaration
```
gabrielittner, line 254: Wouldn't this break using a different screen as parent?

gabrielittner (approved):

codegen-compiler/src/jvmMain/kotlin/com/freeletics/khonshu/codegen/parser/KspParser.kt
gabrielittner: Ah no, because then it's an explicit parameter and no the default value

PR #1391, opened 2026-05-04
Fix `TextResource` arg serialization indices

befrvnk:
Fixes `TextResource` argument serialization for arrays with multiple elements.

`AnyArraySerializer` was encoding every array element with a hardcoded index of `1`, which could break encoding/decoding when multiple arguments were provided. The serializer now uses the actual element index via `forEachIndexed`.


kboyarshinov (approved):
Thank you!

PR #1397, opened 2026-05-09
use more separate bundles to eliminate -android/-desktop dependencies

gabrielittner:
I wanted to see whether we can remove the need of having `-android` and `-desktop` libraries in the version catalog and our build files. This creates more separate bundles and includes the subvariants in them. For lifecycle it's a bit more verbose since that one is combining very different libraries under one group id.

We could add the whole bundle declaration to FGP, it's commonly applicable and also useful for flowredux for example


kboyarshinov (approved):
That looks better! thank you!

PR #1449, opened 2026-07-03
Make DestinationNavigator entry-aware

befrvnk:
### Problem

A destination-scoped navigator currently only knows about the shared `HostNavigator`; it does not know which concrete stack entry it was created for.

This has two drawbacks:

- APIs such as navigation-result registration have to find an entry by route type, which is ambiguous when multiple entries use the same route type.
- A delayed or repeated back action can be triggered after its originating destination is no longer current and accidentally pop a different destination.

This can happen through double-clicks, delayed coroutine callbacks, or similar race-like situations.

### What changed

This PR adds an additive, platform-neutral `DestinationNavigator2` API for generated destinations. It does not replace or change the existing `DestinationNavigator` API.

For each generated destination, codegen provides a scoped `DestinationNavigator2` backed by the destination's exact `StackEntry` and the shared `HostNavigator`:

```kotlin
@Provides
@SingleIn(FooRoute::class)
fun provideDestinationNavigator2(
    hostNavigator: HostNavigator,
    stackEntry: StackEntry<FooRoute>,
): DestinationNavigator2 = DefaultDestinationNavigator2(hostNavigator, stackEntry)
```

`DefaultDestinationNavigator2` is an internal codegen API. App code consumes the `DestinationNavigator2` interface and does not need access to `StackEntry` or `StackEntry.Id`.

### Platform navigation

`DestinationNavigator2` is deliberately platform-neutral. It only exposes `Navigator` operations and the atomic `navigate` block; it does not expose activity or permission navigation APIs.

Generated graphs obtain their `PlatformNavigator` through a provider with an optional legacy `DestinationNavigator` dependency:

```kotlin
@Provides
@SingleIn(FooRoute::class)
fun providePlatformNavigator(
    hostNavigator: HostNavigator,
    @ForScope(FooRoute::class)
    destinationNavigator: DestinationNavigator = DefaultDestinationNavigator(hostNavigator),
): PlatformNavigator = destinationNavigator
```

Metro uses an existing scoped `DestinationNavigator` binding when one is available. Otherwise it uses the internal default, so destinations can use `DestinationNavigator2` without also providing the legacy binding. Generated UI setup calls:

```kotlin
PlatformNavigatorEffect(graph.platformNavigator)
```

This preserves existing `DestinationNavigator` subclasses and their activity-navigation setup while allowing destinations to migrate independently.

### Entry-aware back navigation

Direct back-style operations are guarded by the stack entry that created the navigator:

```kotlin
navigateBack()
navigateUp()
navigateBackTo(...)
```

They only execute while that exact entry is the current/top entry. They no-op when another entry is on top, when another back stack is active, or when the originating entry has already been removed. This prevents a stale navigator from popping a different destination.

Forward and root-changing operations continue to delegate directly to `HostNavigator`:

```kotlin
navigateTo(...)
switchBackStack(...)
showRoot(...)
replaceAllBackStacks(...)
```

The atomic `navigate { ... }` block is guarded at invocation as well and no-ops when the originating entry is not current. Once the block starts, its operations retain `HostNavigator` semantics, so active destinations can still perform transitions such as `navigateBack()` followed by `navigateTo(...)` atomically.

### Navigation results

The existing route-based API remains unchanged:

```kotlin
navigator.registerForNavigationResult<CurrentRoute, Result>()
```

When this API is called on a `DestinationNavigator2` for its own route type, its internal entry lookup resolves to the captured stack entry instead of another entry with the same route type.

A route-free `registerForNavigationResult<Result>()` overload is intentionally not included in this PR and can be introduced separately.

### Migration

A screen-specific navigator can use only the new generated binding:

```kotlin
@Inject
@SingleIn(FooRoute::class)
class FooNavigator(
    navigator: DestinationNavigator2,
) : DestinationNavigator2 by navigator
```

Existing navigator implementations also remain valid and can be migrated destination by destination:

```kotlin
@Inject
@SingleIn(FooRoute::class)
@ForScope(FooRoute::class)
@ContributesBinding(FooRoute::class, binding<DestinationNavigator>())
class LegacyFooNavigator(
    hostNavigator: HostNavigator,
) : DestinationNavigator(hostNavigator)
```

Platform-specific activity or permission navigation should continue to use `DestinationNavigator`/`ActivityNavigator` separately.

### Review feedback incorporated

The final implementation:

- avoids introducing an additional platform-navigation API,
- makes the legacy `DestinationNavigator` binding optional while continuing to use it for platform setup when present,
- allows screen navigators to use only `DestinationNavigator2`,
- scopes the generated `DestinationNavigator2` with `@SingleIn`,
- creates it directly from `HostNavigator` and the current `StackEntry`, and
- keeps navigation-result overload and activity-navigation separation out of this focused change.

### Verification

Verified the full repository build and dependency analysis:

```sh
./gradlew build buildHealth --stacktrace
```

Verified the Android and desktop samples, with `ScreenNavigator` migrated to `DestinationNavigator2` while the other destinations continue using the legacy binding:

```sh
./gradlew -p sample/simple \
  :app:simple-android:assembleDebug \
  :app:simple-desktop:assemble \
  lint \
  --stacktrace
```

Verified the iOS sample for an Apple Silicon simulator as well:

```sh
xcodebuild \
  -project sample/simple/app/simple-ios/iosApp.xcodeproj \
  -scheme iosApp \
  -configuration Debug \
  -sdk iphonesimulator \
  -destination 'generic/platform=iOS Simulator' \
  ARCHS=arm64 \
  ONLY_ACTIVE_ARCH=YES \
  CODE_SIGNING_ALLOWED=NO \
  build
```

Also verified code generation, navigation behavior, and ABI checks:

```sh
./gradlew \
  :codegen-compiler-test:test \
  :navigation:jvmTest \
  :navigation:testAndroidHostTest \
  :navigation:checkKotlinAbi \
  :navigation:checkLegacyAbi \
  :navigation-testing:checkKotlinAbi \
  :navigation-testing:checkLegacyAbi
```


navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/NavigationResults.kt
```diff
@@ -39,6 +53,17 @@ internal fun <T : BaseRoute, O> Navigator.registerForNavigationResult(
     return NavigationResultRequest(key, entry.state, serializer)
 }
 
+@OptIn(com.freeletics.khonshu.navigation.internal.InternalNavigationApi::class)
```
gabrielittner: import

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/NavigationResults.kt
```diff
@@ -39,6 +53,17 @@ internal fun <T : BaseRoute, O> Navigator.registerForNavigationResult(
...
+@PublishedApi
+internal fun <O> DestinationNavigator2.registerForNavigationResult(
+    resultType: String,
+    serializer: KSerializer<O>,
+): NavigationResultRequest<O> {
+    val requestKey = "${stackEntry.id.value}-$resultType"
+    val key = NavigationResultRequest.Key<O>(stackEntry.id, requestKey)
```
gabrielittner: `stackEntry.id` is now present twice in the `Key`

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/NavigationResults.kt
```diff
@@ -27,6 +27,20 @@ public inline fun <reified T : BaseRoute, reified O : Any> Navigator.registerFor
     )
 }
 
+/**
+ * Register for receiving navigation results for this destination entry.
+ *
+ * This overload uses the current entry captured by [DestinationNavigator2] and therefore doesn't require
+ * the route type parameter.
+ */
+@OptIn(com.freeletics.khonshu.navigation.internal.InternalNavigationApi::class)
```
gabrielittner: not needed

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/NavigationResults.kt
```diff
@@ -27,6 +27,20 @@ public inline fun <reified T : BaseRoute, reified O : Any> Navigator.registerFor
     )
 }
 
+/**
+ * Register for receiving navigation results for this destination entry.
+ *
+ * This overload uses the current entry captured by [DestinationNavigator2] and therefore doesn't require
+ * the route type parameter.
```
gabrielittner: This looks like a very weird comment. If I see the code hint for this I need to go search for the other version to see the actual description of the original version. This should just have the same second paragraph as the original method without any changes.

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/DestinationNavigator.kt
```diff
@@ -20,6 +26,28 @@ public abstract class DestinationNavigator(
     }
 }
 
+/**
+ * Exposes the platform-specific navigator implementation for setup in the UI layer.
+ */
+public interface PlatformNavigatorApi {
+    public val platformNavigator: PlatformNavigator
+}
```
gabrielittner: This is a weird API and introduces a lot of indirection. We already had the indirection of `PlatformNavigator` -> `ActivityNavigator`, now we have an extra interface to expose the actual navigator. Also for using the actual navigation APIs this is weird because now you suddenly access them directly anymore on your navigator.

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/DefaultDestinationNavigator.kt
```diff
@@ -0,0 +1,85 @@
...
+    Navigator by hostNavigator {
+    override val platformNavigator: PlatformNavigator
+        get() = this
+
+    /**
+     * Triggers up navigation if the entry that created this navigator is still present.
+     */
+    @OptIn(InternalNavigationTestingApi::class)
+    override fun navigateUp() {
+        ifEntryPresent {
```
gabrielittner: Should be checking whether it's the top entry not whether it's present.

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/DefaultDestinationNavigator.kt
```diff
@@ -0,0 +1,85 @@
...
+    private val hostNavigator: HostNavigator,
+    override val stackEntry: StackEntry<*>,
+) : PlatformNavigator(),
+    DestinationNavigator2,
+    Navigator by hostNavigator {
+    override val platformNavigator: PlatformNavigator
+        get() = this
+
+    /**
+     * Triggers up navigation if the entry that created this navigator is still present.
```
gabrielittner: kdoc on an (effectively) internal class doesn't make sense, also just describes implementation

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/DefaultDestinationNavigator.kt
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * See [HostNavigator.navigate]. The [block] is ignored if the entry that created this navigator is no longer
+     * present.
+     */
+    @OptIn(InternalNavigationTestingApi::class)
+    override fun navigate(block: Navigator.() -> Unit) {
+        ifEntryPresent {
+            navigate(block)
+        }
+    }
```
gabrielittner: Why does this one have an ifEntryPresent wrapper?

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/internal/StackSnapshot.kt
```diff
@@ -70,4 +70,9 @@ public class StackSnapshot internal constructor(
             ?: startStackRootEntry.takeIf { it.id == id }
             ?: throw IllegalStateException("Entry $id not found on back stack")
     }
+
+    internal fun entryForOrNull(id: StackEntry.Id): StackEntry<*>? {
+        return entries.lastOrNull { it.id == id }
+            ?: startStackRootEntry.takeIf { it.id == id }
+    }
```
gabrielittner: Since I left a comment on the usage let's see if we still need it but if it says then the existing `entryFor` should use this internally instead of the logic being duplicated.

codegen-compiler/src/jvmMain/kotlin/com/freeletics/khonshu/codegen/codegen/GraphGenerator.kt
```diff
@@ -155,6 +154,23 @@ internal class GraphGenerator(
...
+                    ),
+                    codeBlock = CodeBlock.builder()
+                        .addStatement(
+                            "return %T(%L, stackEntry)",
+                            defaultDestinationNavigator,
+                            hostNavigator.propertyName,
+                        )
+                        .build(),
+                ),
+            )
```
gabrielittner, line 201: This needs a `@SingleIn` annotation. The current generated code will not work because event call to `destinationNavigator` and every injection side will return a different one, so the one that is observed is separate from the one that is being called.

That also leads to another interesting question for the API design and the use of delegation: Technically the manual `DestinationNavigator2` implementation wouldn't need to be scoped with `@SingleIn` anymore since it's now not stateful anymore. Except when you call registerForResult (including activity variants) in and hold an instance of it then it is suddenly kind of stateful since it holds the request. So maybe the manual class should also always use it? But it's a bit weird. 

There is a second weird part for the delegation and activity result registration. You need to do this before the `PlatformNavigatorEffect` in the UI layer runs. Previously we said you should do that as a property in your `DestinationNavigator` subclass at construction time. Now when you follow that pattern it actually doesn't guarantee anymore that this the right order since a custom subclass with delegation can be instantiated later and separately.

codegen-compiler/src/jvmMain/kotlin/com/freeletics/khonshu/codegen/util/External.kt
```diff
@@ -31,6 +31,8 @@ internal val stackEntry = ClassName("com.freeletics.khonshu.navigation.internal"
 internal val baseRoute = ClassName("com.freeletics.khonshu.navigation", "BaseRoute")
 internal val navRoot = ClassName("com.freeletics.khonshu.navigation", "NavRoot")
 internal val destinationNavigator = ClassName("com.freeletics.khonshu.navigation", "DestinationNavigator")
```
gabrielittner, line 33: Is unused except for calling `destinationNavigator.propertyName` once while other places are now hardcoding `destinationNavigator` as name (the hardcoding is also not great because then the hardcoding potentially needs to be reverted when deleting the original and renaming the 2 variant).

codegen-compiler-test/src/test/kotlin/com/freeletics/khonshu/codegen/codegen/DestinationCodegenTest.kt
```diff
@@ -171,7 +175,7 @@ internal class DestinationCodegenTest {
                 KhonshuTestGraphProvider.provide(entry, snapshot, provider)
               }
 
-              PlatformNavigatorEffect(graph.destinationNavigator)
+              PlatformNavigatorEffect(graph.destinationNavigator.platformNavigator)
```
gabrielittner: This change (always using `DestinationNavigator2` here) effectively makes the whole compatibility part pointless. A graph will always create a `DestinationNavigator2` and will set it up here. A previously created and contributed `DestinationNavigator` will silently stop working. We also can't optionally still support them without having to update something at all usage sites which kind of defeats the purpose of offering both.

gabrielittner (commented):
3 main points from the comments:
- `PlatformNavigatorApi` increases the API surface in a way that is confusing (more types more indirection)
- The usage of interface delegation has some considerations around scoping and register methods that need to be considered
- With the codegen changes the original `DestinationNavigator` is silently broken defeating the point of supporting both

befrvnk:
@gabrielittner  Thanks for the detailed review! I updated the implementation to move further in the direction you suggested.

The main change is that `DestinationNavigator2` is now fully additive and built on top of the existing `DestinationNavigator` binding instead of replacing it.

Generated code still keeps the existing `DestinationNavigator` graph property and still uses it for platform/activity setup:

```kotlin
PlatformNavigatiorEffect(graph.destinationNavigator)
```

Codegen now additionally provides a scoped `DestinationNavigator2`.

```kotlin
@Provides
@SingleIn(FooRoute::class)
fun provideDestinationNavigator(
   @ForScope(FooRoute::class) destinationNavigator: DestinationNavigator,
   stackEntry: StackEntry<FooRoute>,
): DestinationNavigator2 =
   DefaultDestinationNavigator2(destinationNavigator, stackEntry)
```

So existing `DestinationNavigator` subclasses should keep working as before, and the new navigator is only an additive entry-aware API.

I also removed the previous `PlatformNavigatorApi` approach. Instead, I introduced an additive `ActivityNavigatorApi` interface and made the existing `ActivityNavigator` class implement it. `DestinationNavigator2` now extends that interface, and `DefaultDestinationNavigator2` forwards activity/permission APIs to the wrapped `DestinationNavigator`. This lets the new interface support activity and permission navigation without changing the old class-based API.

codegen-compiler/src/jvmMain/kotlin/com/freeletics/khonshu/codegen/codegen/GraphGenerator.kt
kboyarshinov: > Except when you call registerForResult (including activity variants) in and hold an instance of it then it is suddenly kind of stateful since it holds the request. So maybe the manual class should also always use it?

Was also thinking about it but more in direction of detaching Destination navigation from platform specific navigation. Since khonshu is now KMP Android's activity navigation could be moved to additional codegen. `ActivityNavigator` itself can already be just a generated binding. The only caveat as you pointed out is specific ordering requirement of `PlatformNavigatorEffect` and registering for activity results.

I elaborated more [here](https://github.com/freeletics/khonshu/pull/1449#discussion_r3614955067).

docs/codegen/get-started.md
```diff
@@ -92,12 +92,29 @@ a screen is shown in the logged in or logged out state.
...
+`DestinationNavigator` binding.
+
+A class that is contributed as `DestinationNavigator` should not inject `DestinationNavigator2`, because
+`DestinationNavigator2` is built from that `DestinationNavigator` binding.
+
+Because a `DestinationNavigator2` is tied to a concrete stack entry, back-style navigation is guarded:
+`navigateBack`, `navigateUp`, and `navigateBackTo` no-op once the entry that
+created the navigator is no longer on the stack. Forward/root-changing operations like `navigateTo`,
+`switchBackStack`, `showRoot`, and `replaceAllBackStacks` are intentionally not guarded so flows can
+still replace the current entry with another destination.
```
kboyarshinov: I think there's a lot of unnecessary implementation details here for 'getting started' document.

Generally I'd prefer to update the documentation after we finalize implementation and API details.

navigation/src/androidMain/kotlin/com/freeletics/khonshu/navigation/activity/ActivityNavigator.kt
```diff
@@ -19,17 +19,17 @@ import kotlinx.coroutines.flow.receiveAsFlow
  *
  * For this work [ActivityNavigatorEffect] needs to be called.
  */
-public abstract class ActivityNavigator {
+public abstract class ActivityNavigator : ActivityNavigatorApi {
```
kboyarshinov: I think we would not need this if we detach `ActivityNavigator` from `DestinationNavigator2` since most navigators don't use activity navigation. This is breaking change and bigger, but we're already making it backwards compatible for some time.

`ActivityNavigator` can already have `SingleIn` binding in codegen as it's the abstract class without any abstract methods. 

With separate binding Android specific navigation would need to use it for activity navigation and registering for results. It also allows features to separate platform navigation in source sets.

```kotlin
@Inject @SingleIn(PhotoNavDirections::class)
class ScreenNavigator(
    navigator: DestinationNavigator2,
    private val activity: ActivityNavigator
) : DestinationNavigator2 by navigator {
    val photo = activity.registerForActivityResult(TakePicture())
    fun capture() = activity.navigateForResult(photo)
}
```

Now we should keep `ActivityNavigatorEffect` as is since current `DestinationNavigator` relies on it but since it is noop on iOS/Desktop anyway later it could be dropped if we separate codegen per platform, although I'd prefer to avoid it.

This change could also be introduced in separate PR from `DestinationNavigator2` to keep this one focused.

What do you think?

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/NavigationResults.kt
```diff
@@ -27,6 +28,21 @@ public inline fun <reified T : BaseRoute, reified O : Any> Navigator.registerFor
...
 
+/**
+ * Register for receiving navigation results.
+ *
+ * The returned [NavigationResultRequest] has a [NavigationResultRequest.Key]. This `key` should
+ * be passed to different destinations which can then use it to call [deliverNavigationResult]. A
+ * result passed to `deliverNavigationResult` will then be emitted by the `Flow` returned from
+ * [NavigationResultRequest.results].
+ */
+public inline fun <reified O : Any> DestinationNavigator2.registerForNavigationResult(): NavigationResultRequest<O> {
```
kboyarshinov: I think these new overloads could be added in next PR to keep this one focused.

befrvnk:
@kboyarshinov Thanks for the feedback! I've incorporated the review comments into the implementation and updated the PR description to reflect the current state.

kboyarshinov:
@befrvnk Thank you for applying changes. I think this is right direction generally, however, for it to be purely additive there should be a way to use `DestinationNavigator2` without also providing `DestinationNavigator` in the graph. Now it is essentially required since generated destination graph declares it for `PlatformNavigatorEffect`.

For example changing sample to use it would not compile now:

```kotlin
class ScreenNavigator(
    private val route: ScreenRoute,
    val destinationNavigator: DestinationNavigator2,
) : DestinationNavigator2 by destinationNavigator {
```

Using both won't compile either since `navigate` signatures match.

Side note: I think using `DestinationNavigator2` as injected to specific screen Navigator class would be a more desireable use case to avoid ordering issues with activity result registration.

befrvnk:
@kboyarshinov Pushed eight commits since your last review.

[ffbea67](https://github.com/freeletics/khonshu/pull/1449/commits/ffbea67d63bf6d9db637f4e66a44fd936c216520) addresses your point about `DestinationNavigator2` needing to be usable without also providing `DestinationNavigator`: the generated graph now takes the legacy binding as an optional Metro dependency and falls back to a default, and the sample's `ScreenNavigator` is migrated to prove it compiles without one.

The only other design change is [9a129fe](https://github.com/freeletics/khonshu/pull/1449/commits/9a129fea61ee9705499753bee47d6ab526d1d8f8): the entry guard moved from the `DefaultDestinationNavigator2` wrapper into `MultiStackHostNavigator` itself. That removes `DefaultDestinationNavigator2` from the public API and gives `TestHostNavigator` a seam — `DestinationNavigator2`-based navigators weren't testable before ([95b6bec](https://github.com/freeletics/khonshu/pull/1449/commits/95b6bec2b38614c6d9c334502076a9bc92969a22)).

The rest is additive: `isCurrentDestination`, `@OptionalBinding` on the generated provider so it also compiles under `REQUIRE_OPTIONAL_BINDING`, tests for the result-disambiguation case, and docs.

sample/simple/feature/screen/implementation/src/commonMain/kotlin/com/freeletics/sample/screen/ScreenNavigator.kt
```diff
@@ -1,28 +1,24 @@
...
 import com.freeletics.sample.screenwithresult.nav.ScreenWithResultRoute
-import dev.zacsweers.metro.ContributesBinding
 import dev.zacsweers.metro.ForScope
 import dev.zacsweers.metro.Inject
 import dev.zacsweers.metro.SingleIn
-import dev.zacsweers.metro.binding
 
 @Inject
 @ForScope(ScreenRoute::class)
 @SingleIn(ScreenRoute::class)
```
gabrielittner:
```suggestion
```

docs/codegen/get-started.md
```diff
@@ -92,11 +92,16 @@ a screen is shown in the logged in or logged out state.
 
 ## Navigation set up
 
-The integration of Khonshu's Codegen and Navigation libraries also expects a `DestinationNavigator`
-binding to be available. This can be easily achieved by adding `@SingleIn(ExampleScope::class)
-@ForScope(ExampleScope::class) @ContributesBinding(ExampleScope::class, binding<DestinationNavigator>())`
-to a subclass of it. The generated code will automatically take care of setting up
-the navigator by calling `ActivityNavigatorEffect` in the compose UI layer with the navigator.
+Generated destination graphs provide a `DestinationNavigator2` that is scoped to the destination and
+aware of its back stack entry. It can either be injected directly or be delegated to by a screen
```
kboyarshinov: mentioning prior bug is irrelevant

```suggestion
Generated destination graphs provide a `DestinationNavigator2` that is scoped to the destination. It can either be injected directly or be delegated to by a screen
```

docs/navigation/get-started.md
```diff
@@ -124,6 +124,40 @@ navigator.navigateBack()
 navigator.navigateBackTo<MainScreenRoute>(inclusive = false)
 ```
 
+### DestinationNavigator2
+
+`HostNavigator` always operates on the back stack as a whole. That is a problem for navigators that
+belong to a single destination: a back navigation that is triggered from a delayed callback or from
+repeated user input can remove a destination that another screen has put on the back stack in the
+meantime.
```
kboyarshinov: this also describes a prior problem instead of documenting use of `DestinationNavigator2`

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/DestinationNavigator.kt
```diff
@@ -20,6 +21,94 @@ public abstract class DestinationNavigator(
...
+ * the exact back stack entry it was created for. That has two effects:
+ *
+ * - Operations that remove destinations from the back stack ([navigateBack], [navigateUp],
+ *   [navigateBackTo] and [navigate]) are only executed while this navigator's destination is the
+ *   current destination and are ignored otherwise. This makes it safe to trigger back navigation
+ *   from a delayed callback or from repeated user input without accidentally removing a
+ *   destination that was put on the back stack in the meantime.
+ * - [registerForNavigationResult] resolves to this navigator's own destination instead of looking
+ *   up a destination by route type, which is ambiguous when the same route is on the back stack
+ *   more than once.
```
kboyarshinov: this also mentions prior problems of `DestinationNavigator` but maybe it's fine to keep until old one is removed

CHANGELOG.md
```diff
@@ -3,6 +3,28 @@ Change Log
...
+### Navigation
+
+- Added `DestinationNavigator2`, a `Navigator` for a single destination that is aware of the back
+  stack entry it belongs to. `navigateBack`, `navigateUp`, `navigateBackTo` and `navigate` are only
+  executed while its destination is the current destination, so a delayed or repeated back action
+  can not remove a destination that was put on the back stack in the meantime.
+  `registerForNavigationResult` resolves to its own destination instead of looking up a destination
+  by route type, which is ambiguous when the same route is on the back stack more than once.
+  Its `isCurrentDestination` property exposes whether back navigation is currently possible and can
+  be observed from Compose.
```
kboyarshinov: What is the use case for this? just to additionally guard back navigation from Compose?

CHANGELOG.md
```diff
@@ -3,6 +3,28 @@ Change Log
...
+- Generated destination graphs now provide a scoped `DestinationNavigator2`. The existing
+  `DestinationNavigator` continues to work and is still used for the platform navigation set up when
```
kboyarshinov, line 14: worth mentioning that eventually `DestinationNavigator` original will be removed and `DestinationNavigator2` will take its place

docs/navigation/get-started.md
```diff
@@ -124,6 +124,40 @@ navigator.navigateBack()
...
+@SingleIn(DetailScreenRoute::class)
+class DetailScreenNavigator(
+    navigator: DestinationNavigator2,
+) : DestinationNavigator2 by navigator {
+    val result = registerForNavigationResult<DetailScreenRoute, Boolean>()
+}
+```
+
+`DestinationNavigator2` is platform neutral. For activity and permission navigation see
+[Navigating to an Activity](activities.md).
```
kboyarshinov: This needs to be updated to include how to use `DestinationNavigator2` with `ActivityNavigator` + in sample to account for it being android only

kboyarshinov (commented):
@befrvnk Thanks for the adjustments! I think this is ready API wise now. @gabrielittner Would you like to have another look?

CHANGELOG.md
befrvnk: Good question. I couldn't find a real use case, so I removed it in 5e60bb11. Back navigation is already guarded inside the navigator, so nobody needs to check it. And for UI it doesn't add anything: a destination that is covered leaves the composition, so Compose stops its work anyway, and the only one that stays composed is the one under a dialog or bottom sheet, which is still visible and blocked by the modal anyway. The guard itself stays, and `TestHostNavigator.destinationNavigator(isCurrentDestination = false)` still works the same. We can always add the property back if a real need comes up.

kboyarshinov:
@befrvnk I've tried applying it in our app and `PlatformNavigator` API needs some tweaking as `DestinationNavigator` is still necessary for Activity navigation. Just using `PlatformNavigator` doesn't work because `register*` calls have to happen before `PlatformNavigatorEffect` is called. Having `DestinationNavigator` subclass in the graph compiles but doesn't work since results arrive to `DefaultDestinationNavigator` instance created early in the graph.

Could you update the sample with simple location permission request to validate this use case?

Also:
- using optional binding with default value produces a lot of Metro warnings like `@OptionalBinding is redundant in this project as the presence of a default value is sufficient.`
- branch needs to be rebased to resolve conflicts in ABI changes after kotlin update

befrvnk:
Added a simple location permission request to the sample (b9b88c8e). The screen navigator is based on `DestinationNavigator2` and there is no `DestinationNavigator` subclass anymore. It injects the generated `ActivityNavigator` with `@ForScope` and registers for the permission there.

That works because `registerForActivityResult`/`registerForPermissionsResult` can now be called at any time (ea802ba1): requests are kept in an observable list and the effect creates launchers for later registrations without restarting the event collection. One thing to note: Compose doesn't reliably recompose for writes that happen during its own composition pass, so the generated composable now creates the state machine before `PlatformNavigatorEffect`, making constructor registrations visible to the first read (14141756). I also removed the redundant `@OptionalBinding` from the generated provider (01485fbd), which should remove the Metro warnings you saw.

kboyarshinov:
@befrvnk Thank you! I'm not sure about relaxing restrictions on `registerForActivityResult` as it may bring unintended issues and already requires additional logic to handle. Not sure how exactly could it work. What do you think?

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/internal/MultiStackHostNavigator.kt
```diff
@@ -63,6 +64,65 @@ internal class MultiStackHostNavigator(
...
 
+    @InternalNavigationCodegenApi
+    override fun destinationNavigator(entry: StackEntry<*>): DestinationNavigator2 {
+        return EntryNavigator(entry)
+    }
+
+    /**
+     * A [DestinationNavigator2] for a single [entry]. Operations that remove destinations from the
+     * back stack are only executed while [entry] is the current entry.
+     */
```
gabrielittner, line 75: Not a public API and just implements the contract of `DestinationNavigator2`
```suggestion
```

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/internal/MultiStackHostNavigator.kt
```diff
@@ -63,6 +64,65 @@ internal class MultiStackHostNavigator(
...
+    private inner class EntryNavigator(
+        private val entry: StackEntry<*>,
+    ) : DestinationNavigator2,
+        Navigator by this@MultiStackHostNavigator {
```
gabrielittner, line 79: I'd go with an explicit reference here instead of an inner class. This shouldn't need anything private (also see following comments)
```suggestion
    private class EntryNavigator(
        private val entry: StackEntry<*>,
        private val hostNavigator: MultiStackHostNavigator,
    ) : DestinationNavigator2,
        Navigator by hostNavigator {
```

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/internal/MultiStackHostNavigator.kt
```diff
@@ -63,6 +64,65 @@ internal class MultiStackHostNavigator(
...
+        private val isCurrentDestination: Boolean
+            get() = snapshot.value.current.id == entry.id
```
gabrielittner, line 81:
```suggestion
            get() = hostNavigator.snapshot.value.current.id == entry.id
```

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/internal/MultiStackHostNavigator.kt
```diff
@@ -63,6 +64,65 @@ internal class MultiStackHostNavigator(
...
+
+        override fun navigateUp() {
+            if (isCurrentDestination) {
+                stack.popCurrentStack()
```
gabrielittner, line 85: All of these should just call the hostNavigator API instead of duplicating the implementation
```suggestion
                hostNavigator.navigateUp()
```

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/DefaultDestinationNavigator.kt
```diff
@@ -0,0 +1,8 @@
+package com.freeletics.khonshu.navigation
```
gabrielittner, line 1:
```suggestion
package com.freeletics.khonshu.navigation.internal
```

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/DestinationNavigator.kt
```diff
@@ -20,6 +21,78 @@ public abstract class DestinationNavigator(
...
+ * Khonshu's codegen, so this is usually not implemented manually. A navigator class for a specific
+ * screen can delegate to the injected instance:
+ *
+ * ```kotlin
+ * @Inject
+ * @SingleIn(ExampleRoute::class)
+ * class ExampleNavigator(
+ *     navigator: DestinationNavigator2,
+ * ) : DestinationNavigator2 by navigator
+ * ```
```
gabrielittner, line 48: Navigation artifacts should not mention codegen
```suggestion
```

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/DestinationNavigator.kt
```diff
@@ -20,6 +21,78 @@ public abstract class DestinationNavigator(
...
+ *
+ * This is platform neutral and does not offer activity or permission navigation. Use
+ * [DestinationNavigator]/[com.freeletics.khonshu.navigation.activity.ActivityNavigator] for those.
```
gabrielittner, line 51: I don't think we should mention this here. `Navigator` also doesn't say what it can't do
```suggestion
```

navigation/src/commonMain/kotlin/com/freeletics/khonshu/navigation/DestinationNavigator.kt
```diff
@@ -20,6 +21,78 @@ public abstract class DestinationNavigator(
...
+ */
+public interface DestinationNavigator2 : Navigator {
+    /**
+     * Triggers up navigation.
+     *
+     * This is a no-op while this navigator's destination is not the current destination, for
+     * example because it was already removed from the back stack, because another destination was
+     * put on top of it or because another back stack is being shown.
```
gabrielittner, line 59: Seems repetitive on all these methods, maybe just shorter
```suggestion
     * This is a no-op while this navigator's destination is not the current destination.
```

CHANGELOG.md
```diff
@@ -1,15 +1,27 @@
...
+- Added `DestinationNavigator2`, a `Navigator` for a single destination that is aware of the back
+  stack entry it belongs to. `navigateBack`, `navigateUp`, `navigateBackTo` and `navigate` are only
+  executed while its destination is the current destination, so a delayed or repeated back action
+  can not affect another destination. `registerForNavigationResult` resolves to its own destination
+  instead of looking up a destination by route type.
+- Generated destination graphs now provide a scoped `DestinationNavigator2`. The existing
+  `DestinationNavigator` continues to work and is still used for the platform navigation set up when
+  a binding for it exists. Providing one is now optional, so a destination can use only
+  `DestinationNavigator2`. `DestinationNavigator` will be deprecated and removed in a future
+  release, where `DestinationNavigator2` takes its place.
```
gabrielittner, line 17: This is a codegen change not a navigation change.

codegen-compiler/src/jvmMain/kotlin/com/freeletics/khonshu/codegen/codegen/GraphGenerator.kt
```diff
@@ -155,6 +159,44 @@ internal class GraphGenerator(
...
+                        ParameterSpec.builder(hostNavigator.propertyName, hostNavigator).build(),
+                        stackEntryParam,
+                    ),
+                    codeBlock = CodeBlock.builder()
+                        .addStatement(
+                            "return %L.destinationNavigator(stackEntry)",
+                            hostNavigator.propertyName,
+                        )
+                        .build(),
+                    annotations = listOf(singleIn(data.scope)),
```
gabrielittner, line 197: Why does this need a `@SingleIn`? It's stateless, isn't it?

codegen-compiler/src/jvmMain/kotlin/com/freeletics/khonshu/codegen/codegen/DestinationComposableGenerator.kt
```diff
@@ -34,9 +34,12 @@ internal class DestinationComposableGenerator(
...
             .endControlFlow()
             .addStatement("")
-            .addStatement("%M(graph.%L)", platformNavigatorEffect, destinationNavigator.propertyName)
-            .addStatement("")
+            // the content composable is called before the platform navigator effect so that
+            // navigators that are created while it composes (e.g. because they are injected into a
+            // state machine) can register for activity results before the first read
             .addStatement("%L(graph)", composableName)
+            .addStatement("")
+            .addStatement("%M(graph.%L)", platformNavigatorEffect, platformNavigator.propertyName)
```
gabrielittner, line 42: If we have this why do we need the whole dynamic registration logic?

docs/codegen/get-started.md
```diff
@@ -92,11 +92,16 @@ a screen is shown in the logged in or logged out state.
...
+Generated destination graphs provide a `DestinationNavigator2` that is scoped to the destination. It
+can either be injected directly or be delegated to by a screen specific navigator class. See
+[Navigation](../navigation/get-started.md) for what it does.
+
+Providing a `DestinationNavigator` binding is optional. It is only needed for activity and
+permission navigation. When a `@SingleIn(ExampleRoute::class) @ForScope(ExampleRoute::class)`
+binding of `DestinationNavigator` exists, the generated code will use it to set up activity
+navigation by calling `PlatformNavigatorEffect` in the compose UI layer. Note that
+`registerForActivityResult`/`registerForPermissionsResult` still have to be called at construction
+time of that `DestinationNavigator` subclass.
```
gabrielittner, line 104: This should mention how to use `PlatformNavigator`. I wouldn't mention the original `DestinationNavigator` here at all anymore since we're planning to remove it and someone who doesn't know about it shouldn't learn about it.

docs/codegen/get-started.md
```diff
@@ -123,19 +128,18 @@ internal class ExampleStateMachine(
...
-    val repository: ExampleRepository, // a repository that pas provided somewhere in the app
+    val repository: ExampleRepository, // a repository that was provided somewhere in the app
+    val navigator: ExampleNavigator,
 ) : StateMachine<ExampleState, ExampleAction> {
     // ...
 }
 
 @Inject
 // scope the navigator so that everything interacts with the same instance
 @SingleIn(ExampleRoute::class)
```
gabrielittner, line 139: It's stateless now
```suggestion
```

docs/navigation/activities.md
```diff
@@ -19,6 +19,10 @@ will do the required setup to make the `Activity` related navigation actions wor
 It's recommended to make `ActivityNavigator`/`DestinationNavigator` classes specific to
 one screen where they are needed instead of having a global instance.
 
+Since `Activity` navigation is Android only, these classes live in an Android source set. A
+destination that also needs back stack navigation combines them with a `DestinationNavigator2`, see
+[Getting started](get-started.md).
+
```
gabrielittner, line 25:
```suggestion
```

sample/simple/feature/screen/implementation/src/androidMain/kotlin/com/freeletics/sample/screen/ScreenActivityNavigator.kt
```diff
@@ -0,0 +1,36 @@
...
+@ContributesBinding(
+    ScreenRoute::class,
+    binding<LocationPermissionNavigator>(),
+    replaces = [NoOpLocationPermissionNavigator::class],
+)
+class ScreenActivityNavigator(
+    @ForScope(ScreenRoute::class) private val activityNavigator: ActivityNavigator,
+) : LocationPermissionNavigator {
+    // registering can happen at any time, so doing it at construction of a navigator that is
+    // injected into the state machine works as well
```
gabrielittner, line 26: Sounds weird to me. The second part would specifically be when it would need to happen if it can't happen at any point in time, so it doesn't really make sense as an explanation for "can happen at any time".

gabrielittner (commented):
I agree with what Kirill said about dynamic registration. Why can't we enforce before the effect runs?

Haven't looked through the docs yet.

PR #1470, opened 2026-08-18
StackEntryState: update putPlatformValue to verify if all collection types are supported

kboyarshinov:
This fixes crash on Android when trying to store `List` of  `kotlinx.serialization.Serializable` values since `List` itself can either be `@Serializable` or `@Parcelable`.

Test in `StackEntryStateTest` doesn't really verify Android's behaviour since it's on JVM and only tests List serialization. It would pass even without fix since `putPlatformValue` is always `false` on JVM now. New test verifies Android codepath but without building platform classes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,7 @@ Change Log
 
 ### Navigation
 
-- Fixed `StackEntryState` ignoring the `KSerializer` when saving a value that is also `Parcelable`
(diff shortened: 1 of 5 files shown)


zaream (approved):

kboyarshinov:
@zaream I've changed the approach -> instead of preferring serializer `putPlatformValue` on Android now checks for types in collections

zaream (approved):

PR #1490, opened 2026-09-14
Sample: change MainModule to be BindingContainer

kboyarshinov:
Fixes Metro warnings about it


gabrielittner:
I'm not on my laptop so I can't check right now where exactly but the generated code also has this warning somewhere

kboyarshinov:
@gabrielittner Yes, I changed it in https://github.com/freeletics/khonshu/pull/1492

gabrielittner (approved):

PR #1496, opened 2026-09-17
Enable Gradle isolated projects

kboyarshinov:


kboyarshinov:
Can not enable IP yet since state-machine adds js/wasm KMP targets which pull IP incompatible npm gradle plugins. Once module [is removed](https://github.com/freeletics/khonshu/issues/1242) it can be enables as there's no other blockers.

gabrielittner:
I guess you still have a lot of state machine implementations internally?

kboyarshinov:
@gabrielittner yes, there's still several of them, but way less than before.

gabrielittner:
You could have something like this to have flowredux wrappers for them (the updateStateAction is optional), then we could clean things up here.
```kotlin
@ExperimentalCoroutinesApi
inline fun <reified S : Any, reified A : Any> FlowReduxBuilder<S, A>.connectLegacyStateMachine(
    stateMachine: StateMachine<S, A>,
    crossinline updateStateAction: (S) -> A? = { null },
) {
    inState<S> {
        collectWhileInState(stateMachine.state.distinctUntilChanged()) {
            override { it }
        }

        // allows partially migrating state machine or adding new code in flow redux but old
        // state machine needs a way to receive new states
        untilIdentityChanges({ it }) {
            onEnterEffect {
                updateStateAction(snapshot)?.let {
                    stateMachine.dispatch(it)
                }
            }
        }

        onActionEffect<A> {
            stateMachine.dispatch(it)
        }
    }
}
```

kboyarshinov:
@gabrielittner We actually do more or less the same for the biggest one currently for new features and so far migrated some of existing ones too :+1:

